PR #55, opened 2014-05-07
temp test for wmstats problem

ticoann:
@julianbadillo, I added task summary and some other test cases. It needs to be clean up but could be useful to check various things. Could you review?


vlimant:
looks like not relevant anymore

PR #86, opened 2014-12-30
change spec parameter for reqmgr2 update

ticoann:
Hi Julian, this need to be updated when February deploy happens in cmsweb. (reqmgr2 merge). 
https://github.com/dmwm/WMCore/pull/5550, please review.


julianbadillo:
Ok, Thanks @ticoann. I'll merge it when the cmsweb upgrade is over. I guess we need to update the resubmit.py and reqmgr.py scripts?

amaltaro:
@andrewlevin you will also need to adapt these 3 parameters in the relval assignment script once this gets into production

PR #92, opened 2015-02-16
initial attempt to adapt view change

ticoann:
This is not complete patch, need to change the keyword for the return value.

https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Services/WMStats/WMStatsReader.py#L5
following function could be used but preferably just change the underlining key word.
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Services/WMStats/WMStatsReader.py#L33


ticoann:
@julianbadillo, Julian, could you take care of this, Thanks.

PR #93, opened 2015-03-15
A fix for double counting jobs in the tier0 pool and the global pool

lucacopa:
It looks like this is one of the flocking configuration consequences


lucacopa:
@amaltaro @julianbadillo this is tested. It is already in the prod script. Can you merge it?

PR #94, opened 2015-03-19
Tier0 collector changed to vocms007

lucacopa:


lucacopa:
Already working in the prod script.

PR #96, opened 2015-04-07
Unification of scripts

vlimant:
I got the latest versions from Ajit and Andrew, merge them into my repo.
Then developed a set of unified scripts to handle automatic assignment.
This is the initial PR for merging all this.
More development are on the way.

There should be a major review of what is used for what, and most importantly what should be removed.


julianbadillo:
I'm reviewing this, double checking so this doesn't crash what it's working now.
Though some nice comments will be useful once in a while :)

PR #110, opened 2015-06-30
Remove scripts under RelVal/testbed

amaltaro:
As per @ AndrewLevin  request, removing scripts under RelVal/testbed.


julianbadillo:
Ok, let me backup mine into the / folder and I'll merge this

julianbadillo:
@amaltaro what about resubmitTaskChain? I was using it to clone tc's but it looks like the regular resubmit.py works ok.

amaltaro:
I guess the only difference is the loop in line 75 to fix old arguments.

julianbadillo:
This one?
https://github.com/CMSCompOps/WmAgentScripts/blob/master/resubmit.py#L75

RelVal/testbed/resubmitTaskChain.py
```diff
@@ -1,199 +0,0 @@
...
-        elif value != None:
-            schema[key] = value
-#    print "Retrieved schema:"
-#    pprint.pprint(schema)
-#    sys.exit(1)
-    request = deepcopy(schema)
-
-    ### Now changing the parameters according to HG1309
-    x = 1
-    while x <= schema['TaskChain']:
```
amaltaro: Nope, this one! Which is not really important, but helpful when we need to clone an old workflow with old argument names.

julianbadillo:
Then I can copy those options to the resubmit.py, so we can use only one script to clone them all

amaltaro:
Better putting a check for RequestType before, otherwise it fails :-)

PR #111, opened 2015-07-08
Use HTCondor python for Condor Monitoring

lucacopa:
- Replaces all the queries to condor_q and condor_status for the equivalente python condor client way
- Job type is taken from the classAds (CMS_JobType). Logic to decude job type is only a backup now, in case the first is not available.
- Minor changes to add simplicity
- Performance improvement (2 times faster)
- Solves the problem of too many emails :-)


lucacopa:
tested and deployed, I am merging it

PR #112, opened 2015-07-13
Handle jobs not coming from WMAgent

lucacopa:
Some test jobs or dev jobs dont have all the classadds that the monitoring script needs to account them
Ignore jobs like this for now.


lucacopa:
tested and deployed

PR #113, opened 2015-07-29
Tiding up

julianbadillo:
Using HTCondor plugin instead of stdin
PEP8 for some scripts
deletion of uneeded imports.


julianbadillo:
Included all @vlimant changes on a single pull request

julianbadillo:
@vlimant There is some import conflict when we run reqMgrClient outside cmst2, it tries to import utils:
https://github.com/julianbadillo/WmAgentScripts/blob/master/reqMgrClient.py#L12
And utils tries to load the datalock:
https://github.com/julianbadillo/WmAgentScripts/blob/master/utils.py#L150
So it fails to run some basic scripts.
I'm moving the import of utils inside the assignWorkflow
https://github.com/julianbadillo/WmAgentScripts/blob/master/reqMgrClient.py#L685

vlimant:
I'll take a look

PR #114, opened 2015-07-31
Add the original request name to clones

amaltaro:
As discussed with Dima, he only wants to know the original workflow from which a clone was created (not the real first workflow injected into the system for that prepId/etc).

### DIFF
--- a/resubmit.py
+++ b/resubmit.py
@@ -170,8 +170,9 @@ def cloneWorkflow(workflow, user, group, verbose=False, backfill=False):
     """
     # Get info about the workflow to be cloned
     helper = reqMgrClient.retrieveSchema(workflow, reqmgrCouchURL)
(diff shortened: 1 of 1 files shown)


ticoann:
Alan, sorry for late checking on the patch. There is no view for OriginalRequest. I think it should be changed to OriginalRequestName to work with current view.
https://github.com/dmwm/WMCore/blob/master/src/couchapps/ReqMgr/views/byoriginalrequest/map.js#L4

@julianbadillo, Julian could you run the simple test against testbed just check whether parameter gets to couchdb. we don't have to actually run the request. After creation we can just reject it.

amaltaro:
Hmm, I was not aware that there was a view already, sorry :(

ticoann:
It is the view for acdc. But It think we could use that for both.

PR #120, opened 2015-10-16
initial patch to use reqmgr2 api

ticoann:
Still need to add put request and assign etc. This is NOT tested. @julianbadillo, Julian, could you review?

### DIFF
--- a/reqMgrClient.py
+++ b/reqMgrClient.py
@@ -52,8 +52,8 @@ class Workflow:
         else:
             self.outputDatasets = outputdatasetsWorkflow(url, name)
 
-        if 'teams' in self.info and len(self.info['teams']) > 0 :
-            self.team = self.info['teams'][0]
+        if 'Teams' in self.info and len(self.info['Teams']) > 0 :
+            self.team = self.info['Teams'][0]
         else:
             self.team = 'NoTeam'
         if 'FilterEfficiency' in self.info:
@@ -311,14 +311,15 @@ def requestManagerGet(url, request, retries=4):
     """
     conn  =  httplib.HTTPSConnection(url, cert_file = os.getenv('X509_USER_PROXY'),
                                             key_file = os.getenv('X509_USER_PROXY'))
-    r1=conn.request("GET",request)
+    headers = {"Accept": "application/json"}
+    r1=conn.request("GET",request, headers=headers)
     r2=conn.getresponse()
     request = json.loads(r2.read())  
     #try until no exception
     while 'exception' in request and retries > 0:
         conn  =  httplib.HTTPSConnection(url, cert_file = os.getenv('X509_USER_PROXY'),
                                                 key_file = os.getenv('X509_USER_PROXY'))
-        r1=conn.request("GET",request)
+        r1=conn.request("GET",request, headers=headers)
         r2=conn.getresponse()
         request = json.loads(r2.read())
         retries-=1
@@ -352,15 +353,7 @@ def requestManagerPost(url, request, params, head = def_headers, nested=False):
     conn.close()
     return data
 
-def requestManagerPut(url, request, params, head = def_headers):
-    """
-    Performs some operation on ReqMgr through
-    an HTTP PUT method.
-    url: the instance used, i.e. url='cmsweb.cern.ch' 
-    request: the request suffix url for the POST method
-    params: a dict with the PUT parameters
-    head: optional headers param. If not given it takes default value (def_headers)
-    """
+def _put(url, request, params, head = def_headers):
     conn  =  httplib.HTTPSConnection(url, cert_file = os.getenv('X509_USER_PROXY'),
                                     key_file = os.getenv('X509_USER_PROXY'))
     headers = head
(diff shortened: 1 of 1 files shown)


ticoann:
Also, Are these all relevant (used)? Those are putRequest which need to be handled.
https://github.com/CMSCompOps/WmAgentScripts/blob/master/reqMgrClient.py#L878-L983

And these three are needed to be completed changed since those use web interface instead of REST API. (I will do that, I just want to check all three are used currently)

https://github.com/CMSCompOps/WmAgentScripts/blob/master/reqMgrClient.py#L686
https://github.com/CMSCompOps/WmAgentScripts/blob/master/reqMgrClient.py#L985
https://github.com/CMSCompOps/WmAgentScripts/blob/master/reqMgrClient.py#L997

Is there any other modules need to be changed.
I see many files which contains /reqmgr/ url. I am not sure which one is relevant.
Could you point me currently used script which has /reqmgr/ url prefix?

julianbadillo:
Hi @ticoann 
1. About: https://github.com/CMSCompOps/WmAgentScripts/blob/master/reqMgrClient.py#L878-L983
Yes, they are actively used for changing workflow status.
2. About the next three, yes they are also actively used.
3. Maybe these also need to be reviewed:
https://github.com/CMSCompOps/WmAgentScripts/blob/master/utils.py#L1790
https://github.com/CMSCompOps/WmAgentScripts/blob/master/utils.py#L1670-L1768
Though @vlimant knows better.

ticoann:
Thank Julian, I will update the patch.

ticoann:
Hi Julian,
I updated the patch, if you have time could you run the test. If you have any test script I can run it. I can try it too. (Test will be just running the same script to check whether it doesn't break anything - please test against testbed)

We can do other test after following script is updated. (reqmgr2 request need to get in)
https://github.com/CMSCompOps/WmAgentScripts/blob/master/reqMgrClient.py#L686
https://github.com/CMSCompOps/WmAgentScripts/blob/master/reqMgrClient.py#L985
https://github.com/CMSCompOps/WmAgentScripts/blob/master/reqMgrClient.py#L997



> 3. Maybe these also need to be reviewed:
> https://github.com/CMSCompOps/WmAgentScripts/blob/master/utils.py#L1790
> https://github.com/CMSCompOps/WmAgentScripts/blob/master/utils.py#L1670-L1768

These seems to be OK. Are you using deprecated=False option?

Thanks,
Seangchan

vlimant:
deprecated=False most of the time, for certain checks, and members not present in the new workload cache, I had to fall back to using the deprecated workload cache document

julianbadillo:
@ticoann I forgot that we also use this for creating/assignin and changing splitting of requests.
https://github.com/CMSCompOps/WmAgentScripts/blob/master/reqmgr.py
For what I know, it's the same client in the WMCore repo.

reqMgrClient.py
```diff
@@ -371,6 +363,42 @@ def requestManagerPut(url, request, params, head = def_headers):
...
+    """
+    Performs some operation on ReqMgr through
+    an HTTP PUT method.
+    url: the instance used, i.e. url='cmsweb.cern.ch' 
+    request: the request suffix url for the POST method
+    params: a dict with the PUT parameters
+    head: optional headers param. If not given it takes default value (def_headers)
+    """
+    try:
+        return _put(url, request, params, head):
```
julianbadillo: Syntax error here

reqMgrClient.py
```diff
@@ -371,6 +363,42 @@ def requestManagerPut(url, request, params, head = def_headers):
...
+    except httplib.HTTPException as ex:
+        # If we get an HTTPException of 404 means reqmgr2 request
+        if ex.status == 404:
+            # try reqmgr2 call
+            print "%s : reqmgr2 request: %s" % (request, str(ex))
+            cRequest, cParams = _convertToReqMgr2PUTRequest(request, params)
+            return _put(url, cRequest, cParams, head):
```
julianbadillo: Syntax Error

reqMgrClient.py
```diff
@@ -399,8 +427,8 @@ def getWorkflowInfo(url, workflow):
     """
```
julianbadillo: @ticoann When I try to run against testbed, I get "406 Not Acceptable" error.
I used this url: "cmsweb-testbed.cern.ch"
and this workflow: "jbadillo_RunIIWinter15GS-001-T2_CH_CERN_T0Backfill_150527_153801_6417"

ticoann:
Thanks Julian, I updated the syntax error.

reqMgrClient.py
ticoann: Julian, could you give me that commends or script you are using for the testing? It sounds like there is accept header not set for this. Anyway, I think it is easier if I can run the test. Thanks

julianbadillo: import WmAgentScripts.reqMgrClient as req
url='cmsweb-testbed.cern.ch'
wf='jbadillo_RunIIWinter15GS-001-T2_CH_CERN_T0Backfill_150527_153801_6417'
req.getWorkflowInfo(url,wf)

ticoann: Thanks Julian,
Actually I meant all the test you are running, not just this case. Do you have a script for that? Otherwise, I will just create one.

julianbadillo: @ticoann you meant that the commands I posted above actually worked for you?
Because I couldn't make them work, this is my actual test script.
https://github.com/julianbadillo/WmAgentScripts/blob/jbadillo-work/test_reqmgrscripts.py

ticoann: Thanks Julian, the script you point is exactly what I wanted.
One more question, when you run this test do you set the environment using following script?

https://github.com/CMSCompOps/WmAgentScripts/blob/master/setenvscripts.sh

If not, how do you set the environment, more specifically which python version are you using for the test?

julianbadillo: Hi @ticoann 
No, I don't use the setenvscripts.sh, it's pointing to old script location. We may need to update it.
My procedure for running a test is the following:
https://twiki.cern.ch/twiki/bin/view/CMSPublic/CompOpsWorkflowOperationsWMAgentScripts#Setting_Up_Scripts
Basically on vocms049 (or any other agent-machine):
voms-proxy-init -voms cms
export X509_USER_PROXY=/tmp/x509up_uXXXX
source /data/srv/wmagent/current/apps/wmagent/etc/profile.d/init.sh
I don't ask for any specific python version, specifically in vocms049, points to this python:
/data/srv/wmagent/v1.0.8.patch1/sw.pre/slc6_amd64_gcc481/external/python/2.6.8-comp9/bin/python

ticoann:
@julianbadillo, Hi Julian, I updated the patch could you try again?

julianbadillo:
@ticoann The tests worked fine, is it compatible with current production environment?
Should I wait for the next cmsweb upgrade to merge this pull request?

ticoann:
Thanks Julian, it should be compatible with current production. You don't have to wait for the production update. Let me know if you find any problem. There are more things to be changed for assignment and closed out, etc. but I will make separate patch.

julianbadillo:
@ticoann I kept testing this version on my environment.
I had to do a few changes in several scripts all around to adapt. A few notices:
1. When I create an acdc, the information returned by this:
https://github.com/ticoann/WmAgentScripts/blob/reqmgr2_change/reqMgrClient.py#L403
comes without "ProcessingVersion", which is dfferent that what happened with ReqMgr1 (the acdc had the same "ProcessingVersion" than the original, as should be).

2. When a request is still in assigned-approved, the output of the same call may come without "SiteWhitelist", "ProcessingVersion" or "MergedLFNBase" . This is not a big issue but changes our logic sometimes.

julianbadillo:
@ticoann, also ACDC info comes without "ProcessingString".

PR #121, opened 2015-10-22
Keep in Sync

vlimant:
Adding initial hooks for ReReco in Unified


amaltaro:
As discussed via private emails (with Seangchan, JR, MAtteo and Dima), I'm merging it.

PR #123, opened 2015-11-09
Make recovery script work with DBS3 (and PNN stuff)

amaltaro:
In the end it was only 2 lines of change. We're still running a recovery workflow, so don't merge until we get positive results.

Besides that, there are 2 things to be fixed here:
1. handle ReReco with IncludeParents flag on
2. handle ReReco with lumiList

For more info, see
https://cms-logbook.cern.ch/elog/Workflow+processing/22154

### DIFF
--- a/recover/recover-v2.py
+++ b/recover/recover-v2.py
@@ -205,7 +205,7 @@ def getOutputModules(workload, initialTask = None):
     return outputModules
 
 def getFiles(datasetName, runBlacklist, runWhitelist, blockBlacklist,
-             blockWhitelist, dbsUrl):
+             blockWhitelist, dbsUrl, fakeLocation=False):
     """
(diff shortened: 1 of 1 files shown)


amaltaro:
As Julian noticed, we need to run it from a new wmagent release (1.0.11.pre8 or above).
I used vocms009, anyone can use that node until we upgrade production agents.

amaltaro:
Julian, as we discussed, I added a "-f" option that fakes input block location to FNAL and CERN storage elements (only when the block does not have a valid location like MSS or Buffer). This way we do not upload docs to the ACDCServer with empty location (that fails to be acquired).

Also kept some of my logging prints. Feel free to merge it and proceed with additional changes at your will ;)

PR #129, opened 2016-02-26
initial patch to use reqmgr2 api (Do not merge, yet)

ticoann:
There is error reported in assign script

### DIFF
--- a/assignProdTaskChain.py
+++ b/assignProdTaskChain.py
@@ -153,15 +153,6 @@ def assignRequest(url, workflow, team, site, era, procstr, procver, activity, lf
     if verbose:
         print res
 
-def getRequestDict(url, workflow):
-    conn = httplib.HTTPSConnection(url, cert_file=os.getenv(
-        'X509_USER_PROXY'), key_file=os.getenv('X509_USER_PROXY'))
-    r1 = conn.request("GET", '/reqmgr/reqMgr/request?requestName=' + workflow)
-    r2 = conn.getresponse()
-    request = json.loads(r2.read())
-    return request
-
-
 def main():
     url = 'cmsweb.cern.ch'
     # Example: python assignWorkflow.py -w amaltaro_RVZTT_120404_163607_6269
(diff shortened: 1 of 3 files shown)


ticoann:
https://github.com/CMSCompOps/WmAgentScripts/blob/master/changePriorityWorkflow.py#L11

Is this call still used?

vlimant:
submitWorkflow(url, schema, reqmgr2=False) that should be set to True by default, assuming that we'll send any new workflow to reqmgr : that will make it less ambiguous in testing too

vlimant:
+    elif request == "/reqmgr/reqMgr/closeout" or request == "/reqmgr/reqMgr/announce":
+        request = "/reqmgr/reqmgr2/data/request%s" % params['requestName']
+        newParams = {"RequestStatus": "closed-out", "cascade": params["cascade"]}

should be fixed, correct ?

vlimant:
cParams["RequestStatus"] = params["RequestStatus"]
is wrong, correct ?

vlimant:
is the announce cascade really a post instead of a put ?

ticoann:
Thanks Jean-Roch, I fixed the bugs you found. 

> is the announce cascade really a post instead of a put ?

It seemed it was POST call in reqmgr1, but in reqmgr2 it is PUT call.

vlimant:
+    params["Teams"] = teams 

it's supposed to be "Team" = teams[0]

vlimant:
don't bother updating the patch, I am fixing this and will make the PR for testing once all is fixed.

ticoann:
Thanks Jean-Roch, Could you post your patch, then? Since Jen need to test all her scripts as well.

vlimant:
of course I will, working with a functioning patch will be more efficient

ticoann:
Sure, but Jen was testing against this patch. So she has to test again with your patch. With the current fix you found and fixed Jen/Paola should be able to run there test since the test they ran was successful.
However, if it won't take much time to fix all for United script. I guess they can wait.

vlimant:
unified is set. the branch I refered to functions, and can replace the patch which was buggy anyways.
what successful test ? those in https://github.com/dmwm/WMCore/issues/6640 have to restart anyways

PR #134, opened 2016-03-21
Merging unified testbed reqmgr2

vlimant:
includes #133


vlimant:
@amaltaro do you want to take a look ? this is the prod version of unified. and several manual scripts have been tested. we should merge for further development on scripts

Unified/assignor.py
```diff
@@ -246,7 +247,8 @@ def assignor(url ,specific = None, talk=True, options=None):
         print "Allowed",sites_allowed
         if options.primary_aaa:
             sites_allowed = initial_sites_allowed
-            options.useSiteListAsLocation = True
+            #options.useSiteListAsLocation = True
+            options.TrustSitelists = True
```
amaltaro: you may want to implement some mechanism for the PU as well (TrustPUSitelists)

Unified/assignor.py
```diff
@@ -322,6 +325,17 @@ def assignor(url ,specific = None, talk=True, options=None):
         #    sendEmail("sending work to SDSC","%s was assigned to SDSC"% wfo.name, destination=['boj@fnal.gov'])
         
 
+        if False and 'T2_CH_CERN' in parameters['SiteWhitelist']:
```
amaltaro: How about `if 'T2_CH_CERN' not in parameters['SiteWhitelist']:`

Unified/closor.py
```diff
@@ -20,7 +20,13 @@ def spawn_harvesting(url, wfi , in_full):
     all_OK = {}
     requests = []
     outputs = wfi.request['OutputDatasets'] 
-    if 'EnableHarvesting' in wfi.request and wfi.request['EnableHarvesting']:
+    if ('EnableHarvesting' in wfi.request and wfi.request['EnableHarvesting']) or ('DQMConfigCacheID' in wfi.request and wfi.request['DQMConfigCacheID']):
+        if not 'MergedLFNBase' in wfi.request:
+            print "fucked up"
```
amaltaro: Cool!

Unified/assignor.py
vlimant: we'll go about it after merge.

Unified/assignor.py
vlimant: need first to have the HLT in interfill mode and further tuning

Unified/completor.py
```diff
@@ -19,6 +19,9 @@ def forceComplete(url, wfi):
             #sendEmail("force completing","%s is worth force completing\n%s"%( member['RequestName'] , percent_completions))
             print "setting",member['RequestName'],"force-complete"
             reqMgrClient.setWorkflowForceComplete(url, member['RequestName'])
+        elif member['RequestStatus'] in ['acquired','assignment-approved']:
```
amaltaro: you cannot reject workflows in 'acquired', you have to abort it innstead

vlimant: reqMgrClient.invalidateWorkflow takes care of switching from aborted to rejected. something we could have in reqmgr2 to do it transparently.

reqMgrClient.py
```diff
@@ -474,17 +515,32 @@ def outputdatasetsWorkflow(url, workflow):
...
-    datasets = requestManagerGet(url,'/reqmgr/reqMgr/outputDatasetsByRequestName?requestName='+workflow)
+    results = requestManagerGet(url,'/reqmgr2/data/request?name='+workflow)['result']
+    request = results[0][workflow]
+    datasets = []
+    if "OutputDatasets" in request:
+        datasets.extend(request['OutputDatasets'])
+        
+    if "TaskChain" in request:
+        for num in range(request['TaskChain']):
+            if"OutputDatasets" in request["Task%i" % (num+1)]:
```
amaltaro: there is not OutputDatasets key inside a Task (same for Step)

reqMgrClient.py
```diff
@@ -902,73 +932,97 @@ def assignWorkflow(url, workflowname, team, parameters ):
...
 def setWorkflowForceComplete(url, workflowname):
-    """
-    Sets a workflow state to force-complet
-    """
-    params = {"requestName" : workflowname,"status" : "force-complete"}
-    data = requestManagerPut(url,"/reqmgr/reqMgr/request", params)
-    return data
+    return forceCompleteWorkflow(url, workflowname)
 
 def setWorkflowRunning(url, workflowname):
```
amaltaro: Such method should not exist, this transition is made by the agent, not by human.
BTW, "running" state has been deprecated long ago, it's now either "running-open" or "running-closed"

reqMgrClient.py
vlimant: indeed, I let @ticoann to comment on this, it's harmless for the #134

amaltaro:
Ok, I gave up at around 70% of this PR. I've done a dynamic scan and made a few suggestions. A few more general suggestions would be:
* use with() instead of open() when dealing with file objects. This way you guarantee the file will be closed regardless of what happens
* squash ALL these commits into a single one, it's much easier to track/history issues
* :-D needless to say again, if there is no dependency, smaller PRs can be better reviewed and are less bug-prone.

Thanks for updating all these stuff though!

Unified/completor.py
amaltaro: we validate state transition and I think we should keep enforcing that.

vlimant:
thanks for looking.
Each commit covers a specific change, that would count as a PR, then 81 PR might be better to review, but rather impossible to manage on a branch level.
should there be a major change, with one big commit and a few fixes, yes a lone PR is worth it

Unified/completor.py
vlimant: reqMgrClient.invalidateWorkflow takes care of if 'acquired' set 'aborted', else set 'rejected'

PR #135, opened 2016-04-12
The script was modified to correcty assign Taskchain ACDC over merge …

prozober:
The script was modified to:
- Take into account the case where the request type is a resubmission, and we are dealing with a merge task.
- Assign TrustSitelists and TrustPUSitelists values.

### DIFF
--- a/assignProdTaskChain.py
+++ b/assignProdTaskChain.py
@@ -123,13 +123,8 @@ def assignRequest(url, workflow, team, site, era, procstr, procver, activity, lf
               "checkbox" + workflow: "checked"}
     # add xrootd (trustSiteList)
     if trust_site:
-        params['useSiteListAsLocation'] = True
-        
-    # if era is None, leave it out of the json
-    if era is not None:
-        params["AcquisitionEra"] = era
-    if procstr is not None:
-        params["ProcessingString"] = procstr
+        params['TrustSitelists'] = True
+        params['TrustPUSitelists'] = True
 
     # if era is None, leave it out of the json
     if era is not None:
@@ -146,6 +141,8 @@ def assignRequest(url, workflow, team, site, era, procstr, procver, activity, lf
     if verbose:
         pprint(params)
 
+    #sys.exit(1)
+
     # TODO try reqMgr standard
     res = reqMgr.requestManagerPost(url, "/reqmgr/assign/handleAssignmentPage", params, nested=True)
     print 'Assigned workflow:', workflow, 'to site:', site, 'and team', team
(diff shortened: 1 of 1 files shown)


assignProdTaskChain.py
```diff
@@ -146,6 +141,8 @@ def assignRequest(url, workflow, team, site, era, procstr, procver, activity, lf
     if verbose:
         pprint(params)
 
+    #sys.exit(1)
```
amaltaro: you may remove this line :)

assignProdTaskChain.py
```diff
@@ -294,7 +306,24 @@ def main():
...
+
+        elif schema["RequestType"] == "Resubmission":
+            # For resubmission of a merge task inside a taskchain workflow, we cannot provide the acqera and procstring
+            if "Merge" in schema["InitialTaskPath"].split("/")[-1]:
+                acqera = None
+                procstring = None
+            # For another type of task, we need to look for the acqera and procstring information inside the
+            # original workflow
+            else:
+                if any("ProcessingString" in k for k, v in schema.items()):
```
amaltaro: I'd replace it by something like `if "ProcessingString" in schema and "AcquisitionEra" in schema:`

assignProdTaskChain.py
```diff
@@ -294,7 +306,24 @@ def main():
...
+                    procstring = schema["ProcessingString"]
+                elif not procstring:
+                    procstring = findOriginalProcessingString(url, schema["OriginalRequestName"])
+                acqera = schema["AcquisitionEra"]
```
amaltaro: then you move this line to the if just above.

assignProdTaskChain.py
```diff
@@ -154,12 +149,24 @@ def assignRequest(url, workflow, team, site, era, procstr, procver, activity, lf
...
+    conn = httplib.HTTPSConnection(url, cert_file=os.getenv('X509_USER_PROXY'), key_file=os.getenv('X509_USER_PROXY'))
+    urn = "/reqmgr2/data/request/%s" % workflow
+    conn.request("GET", urn, headers=headers)
     r2 = conn.getresponse()
-    request = json.loads(r2.read())
-    return request
+    request = json.loads(r2.read())["result"][0]
+    return request[workflow]
+
+def findOriginalProcessingString(url,workflow):
```
amaltaro: And this method is tricky. You have to return a (or two) dictionary, like:
```
{"AcquisitionEra": {tempSchema['Task1']['TaskName']: tempSchema['Task1']['AcquisitionEra'],
                              tempSchema['Task2']['TaskName']: tempSchema['Task2']['AcquisitionEra'],
                              etc}
 "ProcessingString": {tempSchema['Task1']['TaskName']: tempSchema['Task1']['ProcessingString'],
                              tempSchema['Task2']['TaskName']: tempSchema['Task2']['ProcessingString'],
                              etc}
}
```
and then you modify you return and where you call this function according.

vlimant:
There is something in utils.py (used by Unified) that picks AquisitionEra and ProcessingString up automatically. We can take the chance at using this instead. Where was these info taken from before ? For ACDC and clones, it should be enough to have None for Merges, and nothing specified otherwise, as it's already in the schema of the ACDC.
For assigning plain new taskchain, we would need a way to get this from the schema indeed, or in command line.

amaltaro:
> ACDC and clones, it should be enough to have None for Merges, and nothing specified otherwise, as it's already in the schema of the ACDC.

Unless I misunderstood you, ACDC for TaskChain + Merge task must NOT specify those 2 parameters, None is a different story and will fail assignment.

vlimant:
yes, should not specify it indeed in case of Merge. I was referring to
+                acqera = None
+                procstring = None

amaltaro:
Paola, just to avoid any confusions. NEVER use send AcquisitionEra=None nor ProcessingString=None to reqmgr during assignment. It should either (for resubmission):
a) key/value not be present (for the merge case we fixed yesterday)
b) key/value be present with correct values (for production/processing cases)

vlimant:
Have tested/used this for assigning properly ACDCs ?

PR #136, opened 2016-04-13
add merged scripts

areinsvo:
Added new script, assign.py, which merges the functionality of assignWorkflow.py and assignProdTaskChain.py.

### DIFF
--- a/condor_overview2.py
+++ b/condor_overview2.py
@@ -19,7 +19,7 @@ except ImportError:
     print "source /data/srv/wmagent/current/apps/wmagent/etc/profile.d/init.sh"
     sys.exit(1)
 
-global_pool = 'vocms099.cern.ch'
+global_pool = 'cmsgwms-collector-global.cern.ch'
 
 # production schedd's
 schedds = ["vocms0304.cern.ch",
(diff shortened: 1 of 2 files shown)


condor_overview2.py
```diff
@@ -0,0 +1,365 @@
...
+import optparse
+from random import choice
+try:
+    import htcondor
+except ImportError:
+    print "You do not have a proper environment, please source the following:"
+    print "source /data/srv/wmagent/current/apps/wmagent/etc/profile.d/init.sh"
+    sys.exit(1)
+
+global_pool = 'vocms099.cern.ch'
```
prozober: We need to change this line because the HTCondor collector is going to be migrated (https://hypernews.cern.ch/HyperNews/CMS/get/comp-ops/2929.html). It is better if we do some tests.

areinsvo: I changed the line Paola mentioned above, so that global_pool points to the landb alias 'cmsgwms-collector-global.cern.ch' for monitoring systems. I've tested it and it works as expected.

vlimant:
Could you please go as far as using the siteInfo class in utils.py to get the list of sites in a more dynamic way ?
Can you please report below what are the tests that you have performed for this assign.

vlimant:
please make --lfn mandatory.
we'll find a way to fetch the value from a relevant place in due time : i.e after merging this

PR #137, opened 2016-04-14
Scripts merging

prozober:
The rejectWorkflows.py, rejectAndAbort.py and abortAndClone.py script were merged into the reject.py, it rejects/aborts, and clones workflows in a set of predefined status.


reject.py
```diff
@@ -0,0 +1,89 @@
...
+            group = 'DATAOPS'
+        else:
+            parser.error("Provide the workflow of a file of workflows")
+            sys.exit(1)
+        # name of workflow
+        wfs = [args[0]]
+
+    for wf in wfs:
+        workflow = reqMgrClient.Workflow(wf)
+        if workflow.status in {'assigned' , 'assignment-approved','completed','closed-out'}:
```
vlimant: please use the reqMgrClient.invalidateWorkflow and update your branch

PR #139, opened 2016-05-11
Merging scripts

prozober:
Merging the rejecting scripts(rejectWorkflows.py,abortWorkflows.py,rejectAndClone.py and abortAndClone.p) and the resubmiting ones (resubmit.py and extendWorflow.py)


vlimant:
the rejection does not take care of rejecting the corresponding ACDC if existing. rejector.py does, so let's keep this in mind when we go about having only one share "reject"

PR #142, opened 2016-05-25
Several changes: Avoid the assignment of a task chain wf with a wrong…

prozober:
Several changes were made:

- Avoid the assignment of a task chain workflow with a wrong processing version.

- Set the default values for the resubmit script.

- And improve the reject merging script, now we can set the option to invalidate or not the dataset

### DIFF
--- a/assignProdTaskChain.py
+++ b/assignProdTaskChain.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 """
  +    __author__ = "Paola Rozo"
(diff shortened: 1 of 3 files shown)


vlimant:
looks like this needs to be rebase to the master to avoid conflicts.
https://github.com/CMSCompOps/WmAgentScripts/pull/136 will have to be updated accordingly

prozober:
The rebase is done.

PR #145, opened 2016-05-30
Using the function assignWorkflow of the reqMgrClient to assign task-…

prozober:
Using the function assignWorkflow of the reqMgrClient to assign task-chain workflows.


vlimant:
Paola, do you confirm you have tested and succeeded in assigning wf with this version and version only ?

prozober:
I confirm I've made all the necessary tests to say the script works in every case.

PR #147, opened 2016-06-07
Properly handle reqmgr one response for request creation

amaltaro:
Update required after today's cmsweb upgrade. More details on:
https://hypernews.cern.ch/HyperNews/CMS/get/comp-ops/3024.html

@prozober @areinsvo Matteo (I dont know the user), please test and review


vlimant:
what if the creation failed ? there should be an if ... return None to catch this in a backward compatible way

amaltaro:
Darn it. My local repo was a bit messed up, which is worse now. I'm starting from scratch.

PR #149, opened 2016-06-07
Add option to change memory in reject.py and resubmit.py

areinsvo:
Modified reject.py and resubmit.py to add --memory option to change the memory of the workflow. Tested resubmit.py using backfill jobs to make sure workflow memory is assigned as expected both with and without the option. MaxRSS will be set at assignment to Memory*1024 by the request manager client.


vlimant:
have you checked that cloneWorkflow is not called anywhere with bwl ?

vlimant:
These changes were just washed away with https://github.com/CMSCompOps/WmAgentScripts/pull/151/commits/53540de2b9bcc94aab156df0318a3e30b516602d @areinsvo please make the PR again to re-include these cahnges.

PR #154, opened 2016-06-16
More fixes to assign.py, redo memory option in reject.py and resubmit.py

areinsvo:
Several changes to assign.py, including reverting back to the correct way to determine if we are dealing with a merge task when setting era and proc string. Other changes include setting sites using siteInfo class if --site "t1" or "t2" is given, and allowing the user to pass a comma-separated list of sites.

Added option to set max memory in reject.py and resubmit.py. These should be identical to the changes originally submitted in https://github.com/CMSCompOps/WmAgentScripts/pull/149

### DIFF
--- a/assign.py
+++ b/assign.py
@@ -128,7 +128,6 @@ def assignRequest(url, workflow, team, sites, era, procversion, activity, lfn, p
     if verbose:
         print res
 
-
 def getRequestDict(url, workflow):
     conn = httplib.HTTPSConnection(url, cert_file=os.getenv(
         'X509_USER_PROXY'), key_file=os.getenv('X509_USER_PROXY'))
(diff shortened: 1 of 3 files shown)


vlimant:
@prozober @jenimal can you please double-confirm that this effectively function for several use cases ?

prozober:
I made several tests (with Alan's help) assigning merge and non-merge tasks, and everything works good.

assign.py
```diff
@@ -258,13 +257,15 @@ def main():
...
-        if schema["RequestType"] == "Resubmission" and wf.info["PrepID"].startswith("task") and "Merge" in schema["InitialTaskPath"].split("/")[-2]:
+        if schema["RequestType"] == "Resubmission" and wf.info["PrepID"].startswith("task") and "Merge" in schema["InitialTaskPath"].split("/")[-1]:
             era = None
             procstring = None
 
         # Must use --lfn option, otherwise workflow won't be assigned
         if options.lfn:
             lfn = options.lfn
+        elif "MergedLFNBase" in wf.info:
+            lfn = wf.info['MergedLFNBase']
```
prozober: To avoid any mistake, this value can be found on the original request, am I right? This would be very useful in the ACDC assignment.

jenimal: The best would be to have it read from the origional request and assign 
it the same way, and leave changing it as an option because by default, 
it would needs to be assigned to the same MergedLFNBase as the parent, 
and right now we need to look that up in the parent request each time we 
assign it.

Jen

PR #158, opened 2016-07-04
Memory is set as a float

prozober:
Memory is set as a float in resubmit.py and reject.py


prozober:
Hi @vlimant, have you had the chance to check this?

vlimant:
thanks

PR #163, opened 2016-07-25
The processing string and acquisition era for tasks belonging to Task…

prozober:
The processing string and acquisition era for tasks belonging to TaskChain workflows, are correctly set when we want to create backfills.


vlimant:
looks isolated enough to me

PR #170, opened 2016-08-30
a fix to make things faster and actually get the parent information

vlimant:
up for scrutiny @amaltaro @ticoann @prozober


amaltaro:
The best would be to check whether the original request has `IncludeParents` set to True and then make the proper dbs call with parents.

ticoann:
I think Alan is right, the caller of the listFilesInBlockWithParents has to know whether parent file is needed for processing or not. 

Anyway, updated the WMCore. It will raise DBSReader error if this is called by a block without parents. 
https://github.com/dmwm/WMCore/pull/7129

Thanks,

vlimant:
I am going to merge this in for now for the huge performance gain. It works like this and will refine later on checking on IncludeParents which I think makes sense.

PR #174, opened 2016-09-14
Now we can clone workflows laying on testbed

prozober:
A new option is included, so we can clone workflows laying on testbed.


vlimant:
we discussed further changes to avoid having extra parameters and changes scripts making calls to resubmit. I think that the set of parameters what we should be passing to cloning should be handled differently : for example passing a dictionary along instead of an argument for each. @areinsvo interested in following this up as you touch reject.py recently ?

Issue #175, opened 2016-09-16
Default parameters for assign.py


prozober:
The assign.py script should be able to read the xrood and secondary_xrood parameters from the original workflow. Do you think we need to include something else?
@areinsvo @vlimant

vlimant:
yes, I think we are done here.

PR #178, opened 2016-09-16
Parameters we need to take from the original request.

prozober:
We need to take xrootd and secondary_xrootd from the original request. Also the MergedLFNBase was not being read from the original request but from the request to be assigned, that did not make sense. #175

### DIFF
--- a/assign.py
+++ b/assign.py
@@ -235,14 +235,7 @@ def main():
             era = None
             procstring = None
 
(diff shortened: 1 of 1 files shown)


vlimant:
Is it a problem that we cannot unset aaa at all ? I think it is

vlimant:
in #180

PR #180, opened 2016-09-29
Assign from acdc

vlimant:
@prozober as discussed, this pulls in the sites from the acdc doc.

@amaltaro @ticoann applies the memory in assignment is needed.

checks the sites memory/core requirements for overlap with the site whitelist.

to be tested. @prozober just shoot me a chat when there is an acdc to be assigned so that I can try it out.

### DIFF
--- a/assign.py
+++ b/assign.py
@@ -38,7 +38,7 @@ def getRandomDiskSite(site=None):
         s += "_Disk"
     return s
 
-def assignRequest(url, workflow, team, sites, era, procversion, activity, lfn, procstring, trust_site=False, replica=False, verbose=False, taskchain=False, trust_secondary_site=False):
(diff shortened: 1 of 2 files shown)


assign.py
```diff
@@ -227,21 +236,21 @@ def main():
...
             era = None
             procstring = None
 
         # Must use --lfn option, otherwise workflow won't be assigned
         if options.lfn:
             lfn = options.lfn
-        elif "MergedLFNBase" in wf.info:
-            lfn = wf.info['MergedLFNBase']
+        elif "MergedLFNBase" in wf_info:
+            lfn = wf_info['MergedLFNBase']
```
prozober: lfn needs to be taken from the original request. This line is reading the MergedLFNBase value from the request to be assigned, makes no sense.

PR #182, opened 2016-09-30
Resubmit to testbed

vlimant:
supersede #174 
@prozober give it a try
@amaltaro please take a look for obvious things

### DIFF
--- a/reqMgrClient.py
+++ b/reqMgrClient.py
@@ -451,6 +451,8 @@ def getWorkflowInfo(url, workflow):
     """
     Retrieves workflow information
     """
+    request = requestManagerGet(url,'/reqmgr2/data/request?name='+workflow)
+    return request['result'][0][workflow]
     retries=10000
     while retries:
         try:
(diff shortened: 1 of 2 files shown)


amaltaro:
I haven't seen the `ConfigCouchUrl` being explicitly set. But if it worked for you, then it works.

prozober:
I did a couple of tests, everything is OK.

amaltaro:
As mentioned in another thread, just use this mechanism with care (not as standard procedure) because it breaks injection to DBS and block closure in TMDB (meaning the agent will infinitely try to inject blocks produced).

vlimant:
@amaltaro not knowing of the restriction on input data, I think these changes are actually more dangerous than anything else, and I will just nuke the PR and let Federica inject use testbed specific instructions.
BTW : for unified-testbed, we are essentially doomed because of this limitation

amaltaro:
I don't know how the unified-testbed validation happens, but if you can have a set of templates or requests that you always clone from, we can easily migrate all the inputs/PU into the DBS testbed instance.

Of course, if every day you get a random workflow to clone from and process in testbed, then yes, we're all screwed because we share the same stable testbed agent.

PR #186, opened 2016-10-26
makeACDC compatible with reqmgr2

vlimant:
Now passing through a method from unified and reqMgrClient.py
to switch to making ACDC in reqmgr2, either flip the switch in reqMgrClient.submitWorkflow or execute
export UNIFIED_SUBMIT=reqmgr2

@prozober I was able to use it for a couple of ACDC, which I could assign with the recent changes in master.


vlimant:
./makeACDC.py -f cerminar_Run2016G-v1-DoubleMuonLowMass-23Sep2016_8020_160926_173156_2381 -t DataProcessing
works for example

amaltaro:
Jean-Roch, I thought you were not able to work with reqmgr2 ACDCs and we had to fix something on the server side. Can you please give us an update? Is it known not to work with TaskChain ACDCs for instance?

vlimant:
I implemented to use the ancestors value to assign AcDc so that it does not rely on the content of the AcDc schema. I could not find a reason it would not work.
@ticoann told me he still wants to have the proper values added to the acdc workload, but as far as ops is concerned, I think there is no more show-stopper. to be confirmed in the coming days

PR #195, opened 2016-11-26
Allow resizing of jobs down to one core.

bbockelm:
This allows any multi-core job to be resized down to a single-core one.  GlideinWMS will automatically determine if the single-core version of this job fits in the remaining pilot lifetime.

@vlimant - note that, with this, you cannot do performance tuning as the performance is a function of the number of cores.

### DIFF
--- a/Unified/go_condor.py
+++ b/Unified/go_condor.py
@@ -135,7 +135,7 @@ def makeResizeAd():
     anAd["GridResource"] = "condor localhost localhost"
     anAd["TargetUniverse"] = 5
     anAd['Name'] = "Enable job resizing"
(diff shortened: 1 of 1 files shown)


vlimant:
@bbockelm we were chatting about this, I think that we need to have most parameters changeable one way or the other. either directly in the expr, or by having go_condor replace values from initial values of the workflow

bbockelm:
Yes - we'd need to rename a few of the expressions that get changed by other routes (i.e., `RequestMemory` becomes something like `OriginalRequestMemory` or something).  Unfortunately, I don't know how to make a version compatible with both old and new WMAgents.

We might have to wait until Alan updates all the agents, then update the routes.

bbockelm:
Closing.  We went in a different direction.

PR #197, opened 2016-12-18
Start overhaul of overflow routes.

bbockelm:
Includes code for a single route for all overflow jobs, but that isn't
enabled.  For now, just combine sufficiently similar routes.

Reduces the amount of work the JobRouter has to do by about 2-3x.  Once we get the 'master overflow rule' working, it'll be an additional 2-3x (about an order of magnitude combined).


vlimant:
so you have tested this on several schedds ? which ones ?

bbockelm:
This is on vocms0130 and vocms0310.

bbockelm:
Deployed everywhere.

vlimant:
@bbockelm this is crapping out for certain rules. We have to review and roll out if need be.

PR #198, opened 2017-01-03
Master route v2

bbockelm:
Second attempt at creating one route per workflow.

The big change from last time is that we now obey the per-workflow site whitelist.  This means we have to create one route per unique whitelist.  This results in a modest number of routes - about 10 to 20 - but each job is routed only once.


bbockelm:
With JR, tested this on `vocms0310`.  All produced routes looked correct.

PR #210, opened 2017-02-28
recoverMissingLumis and assign changes

prozober:
Two changes:
1. recoverMissinLumis.py: A change in the nodes' name for the blockLocations. 
2. assign.py: If xrootd and secondary xrootd variables are not given, we are going to assign "False" by default, or the values from the original request (if not null).

### DIFF
--- a/assign.py
+++ b/assign.py
@@ -152,6 +152,8 @@ def main():
     sites = []
     specialStr = ''
     taskchain = False
+    xrootd= False
+    secondary_xrootd= False
 
     SI = siteInfo()
     getRandomDiskSite.T1 = SI.sites_T1s
@@ -272,6 +274,17 @@ def main():
             else:
                 procversion = wf_info["ProcessingVersion"]
 
+        # reading xrootd and secondary_xrootd values
+        if options.xrootd:
+            xrootd = options.xrootd
+        elif original_wf:
+            xrootd= original_wf.request["TrustSitelists"]
+
+        if options.secondary_xrootd:
+            secondary_xrootd = options.secondary_xrootd
+        elif original_wf:
+            secondary_xrootd= original_wf.request["TrustPUSitelists"]
+
         # Check for output dataset existence, and abort if output datasets already exist!
         # Don't perform this check for ACDC's
         datasets = schema["OutputDatasets"]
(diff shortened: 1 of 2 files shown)


recoverMissingLumis.py
```diff
@@ -266,13 +266,14 @@ def getFiles(datasetName, runBlacklist, runWhitelist, blockBlacklist,
             for replica in replicaInfo["phedex"]["block"][0]["replica"]:
                 node = replica["node"]
                 cmsSites = siteDB.PNNtoPSN(node)
+                blockLocations.add(node.replace("_MSS","_Disk"))
```
vlimant: @amaltaro anything that is on tape would appear on disk with this, isn't that a bit risky ? shouldn't we just drop the block location if it is on tape ?

assign.py
```diff
@@ -272,6 +274,17 @@ def main():
             else:
                 procversion = wf_info["ProcessingVersion"]
 
+        # reading xrootd and secondary_xrootd values
```
vlimant: the issue with doing so @prozober is that you have no way at all to tell "trust_site=False" for a recovery that was ran with trust_site=True. you would need a tri-valued --xrootd option that can take
--xrootd ==> True
--xrootd False ==> False
or a --no_xrootd and --no_secondary_xrootd

recoverMissingLumis.py
amaltaro: I don't remember why I added this line at that day, but I agree it's risky and not correct. We can either skip it if MSS or add to the blockLocations (which AFAIK will be the same as dropped, since we don't map PSN to MSS PNN).

assign.py
prozober: I made the changes. What do you think about it?

assign.py
```diff
@@ -275,12 +279,12 @@ def main():
                 procversion = wf_info["ProcessingVersion"]
 
         # reading xrootd and secondary_xrootd values
-        if options.xrootd:
+        if options.xrootd is not None:
```
vlimant: have you checked that the "no option" is options.xrootd == None ?

prozober: I checked the--no_secondary_xrootd option for some running ACDCs, of course, without performing the assignment.
In the case of --no_xrootd, I haven't found an ACDC where the original request has been assigned over xrootd. But, as they function the same way, I see no problem.

recoverMissingLumis.py
vlimant: so in the end there is no change in recoverMissingLumis.py ? wasn't there something broken recently ?

PR #214, opened 2017-03-09
Tune original numbers not derived values.

bbockelm:


vlimant:
This would work only for jobs that have had the resizing change, but not for the others. 
Is there a way to add if Memory is expr change OriginalMemory else Memory ?

bbockelm:
Looking at the global pool with this query:
```
condor_q -const '(WMCore_ResizeJob is undefined) && (WMAgent_RequestName isnt undefined)' -g
```
It seems there are only two agents that have not-upgraded jobs.  One is draining and the other is the backup T0 agent.  Neither should be affected by this change.

So, I think we're good to go?

PR #219, opened 2017-04-09
Include router ad for IO tuning.

bbockelm:
Produces an router ad for providing an IO tuning for each job.


vlimant:
done with 0b8bdadca153574ec47d787c4640a295fcac9cc5 however @bbockelm we would need a rule to set the IOToken

bbockelm:
We cannot yet set IOToken -- if we set it before SI rolls out corresponding changes, jobs won't start up.

PR #221, opened 2017-04-28
Added clone API; fixed some wf submission code

amaltaro:
This PR is about adding the clone API, here:
https://github.com/CMSCompOps/WmAgentScripts/compare/master...amaltaro:clone-api?expand=1#diff-3fc98fbeeaadf6df0cfbc50d59551ffeR1120

one can simply provide the request name in the URI and that workflow gets cloned. Or one can provide a encoded python dictionary to be used to override arguments. This also works for overriding only one or two arguments inside a task/step.

Additional changes are:
* use reqmgr2 request api to retrieve the request json
* dropped access to the couchdb workfload_cache data (the same data as above)
* dropped the WMWorkload helper object from some places (need to be well tested). Can someone please clarify why it has to get the workload instance instead of just using the json request?

We still need to distinguish plain clones (no override) from real workflow retry (overriding args, at least ProcVersion). This way we can better use resources. This needs to be addressed ASAP to avoid issues with the upcoming cmsweb upgrade.

Things to follow up/fix (not me :)):
* fix this code I marked as broken (getdataTaskChains.py)
* fix extendWorkflow.py and resubmitUnprocessedBlocks.py for two reasons, that I could see. First, it calls reqmgr several times when it can make a single call. Second, it uses the wmworkload helper as well.

FYI @ticoann 
@prozober can you please test these changes? If you prefer, let me know how to run these tests and on which workflows and I can try it myself.

### DIFF
--- a/resubmit.py
+++ b/resubmit.py
@@ -19,13 +19,14 @@
     This script depends on WMCore code, so WMAgent environment,libraries and voms proxy need to be loaded before running it.
 """
 import os
-import datetime
-import dbs3Client
 import pwd
-import sys
 import re
+import sys
 from optparse import OptionParser
 from pprint import pprint
+
+import dbs3Client
+
 try:
     import reqMgrClient
 except:
@@ -38,7 +39,7 @@ reqmgrCouchURL = "https://cmsweb.cern.ch/couchdb/reqmgr_workload_cache"
 DELTA_EVENTS = 1000
 DELTA_LUMIS = 200
 
-def modifySchema(helper, workflow, user, group, cache, events, firstLumi, backfill=False, memory=None, timeperevent=None):
+def modifySchema(cache, workflow, user, group, events, firstLumi, backfill=False, memory=None, timeperevent=None):
     """
     Adapts schema to right parameters.
     If the original workflow points to DBS2, DBS3 URL is fixed instead.
@@ -49,33 +50,30 @@ def modifySchema(helper, workflow, user, group, cache, events, firstLumi, backfi
     paramBlacklist = ['BlockCloseMaxEvents', 'BlockCloseMaxFiles', 'BlockCloseMaxSize', 'BlockCloseMaxWaitTime',
                       'CouchWorkloadDBName', 'CustodialGroup', 'CustodialSubType', 'Dashboard',
                       'GracePeriod', 'HardTimeout', 'InitialPriority', 'inputMode', 'MaxMergeEvents', 'MaxMergeSize',
-                      'MaxRSS', 'MaxVSize', 'MinMergeSize', 'NonCustodialGroup', 'NonCustodialSubType',
+                      'MaxRSS', 'MaxVSize', 'MinMergeSize', 'NonCustodialGroup','NonCustodialSubType',
                       'OutputDatasets', 'ReqMgr2Only', 'RequestDate' 'RequestorDN', 'RequestName', 'RequestStatus',
                       'RequestTransition', 'RequestWorkflow', 'SiteWhitelist', 'SoftTimeout', 'SoftwareVersions',
                       'SubscriptionPriority', 'Team', 'timeStamp', 'TrustSitelists', 'TrustPUSitelists',
                       'TotalEstimatedJobs', 'TotalInputEvents', 'TotalInputLumis', 'TotalInputFiles']
 
     result = {}
-    for key, value in helper.data.request.schema.dictionary_whole_tree_().items():
-        if key in paramBlacklist or value in ([], {}, None, ''):
+    for k, v in cache.iteritems():
+        if k in paramBlacklist or v in ([], {}, None, ''):
             continue
         else:
-            result[key] = value
+            result[k] = v
 
-        # Now apply the specific tweaks
-        if key == 'Memory' and memory:
-            result['Memory'] = memory
-        if key == 'TimePerEvent' and timeperevent:
-            result['TimePerEvent'] = timeperevent
-        elif key == 'Requestor':
-            result['Requestor'] = user
-        elif key == 'Group':
-            result['Group'] = group
-        elif key == 'MergedLFNBase':
-            result['MergedLFNBase'] = helper.getMergedLFNBase()
-        # and required for cross-cloning between prod and testbed
-        elif key == 'CouchURL':
-            result['ConfigCacheUrl'] = result.pop('CouchURL')
+    if memory:
+        result['Memory'] = memory
+    if timeperevent:
+        result['TimePerEvent'] = timeperevent
+    if user:
+        result['Requestor'] = user
+    if group:
+        result['Group'] = group
+    # and required for cross-cloning between prod and testbed
+    if result.get('CouchURL'):
+        result['ConfigCacheUrl'] = result.pop('CouchURL')
 
     #if we are extending the workflow
     if events:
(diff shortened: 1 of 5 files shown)


resubmit.py
```diff
@@ -251,9 +209,9 @@ def extendWorkflow(workflow, user, group, verbose=False, events=None, firstlumi=
     firstlumi = int(firstlumi)
 
     # Get info about the workflow to be cloned
-    helper = reqMgrClient.retrieveSchema(workflow)
-    schema = modifySchema(helper, workflow, user, group, None, events, firstlumi, None)
-    schema['OriginalRequestName'] = workflow
+    cache = reqMgrClient.getWorkflowInfo(url, workflow)
```
vlimant: @amaltaro I think we have been back and forth a couple of times on using the Info instead of the cache because of missing items in the "info". Can you recall adding everything that had been missing ?
@prozober if I recall correctly, some of the missing parameters were silently not making things fail, but the clone would not have the expected parameters, and use some default values instead.

amaltaro: Maybe it was the merge* keys? But it should be present in all requests for several months now

amaltaro:
Paola, did you manage to test it?

prozober:
Sorry I took so long, I was off work.
I couldn't recall what were the parameters the resubmit script was taking by default using Info, the only thing is they were different from the merge keys.
While I try to recall, I better do some tests. @amaltaro, do we have some workflows I can play with in testbed? There is nothing to kill and clone in production right now.

prozober:
Mayday Alan!! I am trying to clone this workflow as a backfill https://cmsweb.cern.ch/reqmgr2/data/request?name=pdmvserv_EGM-PhaseISpring17DR-00020_00006_v1__170425_161931_4978.

I run 
`python resubmit.py --backfill -a clone pdmvserv_EGM-PhaseISpring17DR-00020_00006_v1__170425_161931_4978
`
And I get 

```
<h2>400 Bad Request</h2>
        <p>Invalid spec parameter value: WMSpecFactoryException
Message: There are unknown/unsupported arguments in your request spec: [u'DN', u'AutoApproveSubscriptionSites', u'NonCustodialSites', u'Teams', u'OutputModulesLFNBases', u'AllowOpportunistic', u'InputDatasets', u'_id', u'RequestNumEvents']
	ModuleName : WMCore.WMSpec.WMWorkloadTools
	MethodName : validateUnknownArgs
	ClassInstance : None
	FileName : /data/srv/beHG1705h/sw/slc6_amd64_gcc493/cms/reqmgr2/1.1.3_cmsweb.patch1/lib/python2.7/site-packages/WMCore/WMSpec/WMWorkloadTools.py
	ClassName : None
	LineNumber : 306
	ErrorNr : 0

Traceback: 
```

resubmit.py
```diff
@@ -49,33 +50,30 @@ def modifySchema(helper, workflow, user, group, cache, events, firstLumi, backfi
     paramBlacklist = ['BlockCloseMaxEvents', 'BlockCloseMaxFiles', 'BlockCloseMaxSize', 'BlockCloseMaxWaitTime',
```
vlimant: a few more entries need to be added in this list.
DN, AutoApproveSubscriptionSites, NonCustodialSites, ...

amaltaro:
Feel free to play with any testbed requests under "HG1703_Validation" campaign.
About solving this issue, Jean-Roch's suggestion is correct. 

However, I'd like to see Ops using the clone API for most of the use cases (if not all). Instead of doing all this juggling with retrieving/removing/updating/injecting arguments.

resubmit.py
prozober: InputDatasets, RequestNumEvents, a bad idea, isn't it?

vlimant: the parameters seem to depend on the what type of request is being cloned, which makes sense, but it's breaking the resubmit code.
N.B. at least one of InputDatasets or RequestNumEvents should be allowed, so there is something fishy @amaltaro

vlimant: InputDatasets mind the **s**, so it's fine

amaltaro: Yep, InputDatasetS was removed. RequestNumEvents sort of depends on the request type. But in the end it relies on whether there is InputDataset or not.

prozober:
When we try to clone a taskchain we are getting:

`python resubmit.py --backfill -a clone pdmvserv_task_TOP-RunIISummer15wmLHEGS-00202__v1_T_170503_124441_8443`

```
  File "resubmit.py", line 312, in <module>
    main()
  File "resubmit.py", line 303, in main
    wf, user, options.group, options.verbose, options.backfill, options.testbed, memory,timeperevent,bwl=options.bwl)
  File "resubmit.py", line 138, in cloneWorkflow
    schema = modifySchema(cache, workflow, user, group, None, None, backfill, memory, timeperevent)
  File "resubmit.py", line 110, in modifySchema
    if "backfill" not in result["AcquisitionEra"].lower():
AttributeError: 'dict' object has no attribute 'lower'
```

amaltaro:
There are a few arguments that can be a python dict. Especially for TaskChain and StepChain.
Thus you have to iterate over the dictionary.
Don't forget updating the task level AcqEra, since that has precedence over the top level.

resubmit.py
vlimant: looks like RequestNumEvents in not in that reqMgrClient.getWorkflowInfo
this is making things fallback to 1000 event that is the default.

amaltaro: nahh, this api
https://github.com/amaltaro/WmAgentScripts/blob/da9422cdeaa1af81cde15535b21d7c9a0093b777/reqMgrClient.py#L458
must yield that key/value.

vlimant:
based on 

Message: There are unknown/unsupported arguments in your request spec: [u'DN', u'AutoApproveSubscriptionSites', u'NonCustodialSites', u'Teams', u'OutputModulesLFNBases', u'AllowOpportunistic', u'InputDatasets', u'_id', u'RequestNumEvents']

 for a request that has InputDataset, RequestNumEvents should be dropped, other it should be kept.
That include stuff in the nested Task dict ; I don't think there is anything in that PR about that. We will see with time and further issue
@amaltaro can you rework that PR ? with dropping 'DN', u'AutoApproveSubscriptionSites', u'NonCustodialSites', u'Teams', u'OutputModulesLFNBases', u'AllowOpportunistic', u'InputDatasets', u'_id', 

 and the conditional drop according to what is expected on the reqmgr2 side ?

amaltaro:
I've just added a commit addressing Jean-Roch's comments. Not tested again :-)
And to avoid all these problems, we should migrate to the clone API in the next production upgrade (June 13), it's going to have several fixes/improvements and should be fully functional. However, you guys need to start working on these scripts and identify everything that could go through the clone api. As discussed in a GH issue, the API will automatically increment the ProcessingVErsion.

amaltaro:
@prozober can you test it please?

prozober:
Alan, I got this when I try to clone a workflow.

```
<h2>400 Bad Request</h2>
        <p>Invalid spec parameter value: WMSpecFactoryException
Message: There are unknown/unsupported arguments in your request spec: [u'_id']
	ModuleName : WMCore.WMSpec.WMWorkloadTools
	MethodName : validateUnknownArgs
	ClassInstance : None
	FileName : /data/srv/beHG1705h/sw/slc6_amd64_gcc493/cms/reqmgr2/1.1.3_cmsweb.patch1/lib/python2.7/site-packages/WMCore/WMSpec/WMWorkloadTools.py
	ClassName : None
	LineNumber : 306
	ErrorNr : 0

Traceback: 
```

amaltaro:
Sorry, Paola, I missed that key. Patch is updated, can you try again please.

prozober:
I cloned the same workflow, and checked all the parameters:  everything seems to be OK. Thanks Alan!

vlimant:
so, good to merge ?

prozober:
Yes, please.

PR #231, opened 2017-06-08
Hook hold release

vlimant:
@bbockelm does this look OK ? any suggestion ?

### DIFF
--- a/Unified/go_condor.py
+++ b/Unified/go_condor.py
@@ -11,6 +11,41 @@ import htcondor
 from collections import defaultdict
 
 
+def makeHoldAds(config):
+    """
+    Create a set of rules to hold a task from matching
+    """
+    for task,where in config.get('hold',{}).items():
+        # task is the task name
+        # where is either an empty list=all sites, or a list of sites (not implemented)
+        anAd = classad.ClassAd()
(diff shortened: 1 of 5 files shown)


Unified/go_condor.py
```diff
@@ -11,6 +11,41 @@
...
+    """
+    Create a set of rules to hold a task from matching
+    """
+    for task,where in config.get('hold',{}).items():
+        # task is the task name
+        # where is either an empty list=all sites, or a list of sites (not implemented)
+        anAd = classad.ClassAd()
+        anAd["Name"] = str("Holding task %s from %s"%(task, where))
+        anAd["GridResource"] = "condor localhost localhost"
+        anAd["TargetUniverse"] = 5
```
amaltaro: Is it going to set the condor JobStatus to 5 (Held)? If so, beware held jobs are considered an error status in WMCore

vlimant:
5 is the universe, not the job status.
we'd use a site whitelist to an inexistant site to prevent jobs from matching

Unified/go_condor.py
bbockelm: Nope, this is a job's `universe` (vanilla).  Should be a no-op.

PR #237, opened 2017-08-29
Script to create a harvesting-only workflow from any other wf

amaltaro:
Has a few aesthetic last minute changes, so you might want to test it before merging (only functionality, logic should be Ok).

There is a need to rerun harvesting on from this workflow:
 zhenhu_RVCMSSW_9_2_8VBFHToBB_M125_Pow_py8_Evt_13UP17__HS_170809_220847_85

see workflow team thread with subject "9_2_9_vs_9_2_8 relmon stuck?"

### DIFF
--- a/injectHarvest.py
+++ b/injectHarvest.py
@@ -1,0 +1,137 @@
+"""
+Pre-requisites:
+ 1. a valid proxy in your X509_USER_PROXY variable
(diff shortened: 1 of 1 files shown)


vlimant:
could this be using reqMgrClient.py to manipulate the request ?

PR #242, opened 2017-10-17
Skip the factor of 5 for lumi check in PhaseII if only processing 1 lumi per job

thongonary:
Tested locally and assigned 4 staging workflows.


vlimant:
this does not constitute an improvement of the code.

PR #249, opened 2017-11-13
Agent plotting

scarletnorberg:
Attempting to do a pr for the plotting script and the tracking json file.

### DIFF
--- a/component_down.py
+++ b/component_down.py
@@ -28,6 +28,8 @@ for component in os.listdir(base_dir):
         print timestamp,time.asctime( restart_date )
         restarts['data'][component].append( restart_time )
 
-open('/afs/cern.ch/user/c/cmst1/www/Automatic_Agent_json/%s.restart.json'% agent,'w').write( json.dumps( restarts, indent=2) )
+open('/data/srv/wmagent/current/bin/%s.restart.json'% agent,'w').write( json.dumps( restarts, indent=2) )
+os.system("xrdcp *.json root://eoscms.cern.ch//eos/cms/store/logs/prod/recent/agent_crashes")
+#open('/eos/cms/store/logs/prod/recent%s.restart.json'% agent,'w').write( json.dumps( restarts, indent=2) )
 #want to add to eos /eos/project/c/cms-unified-logs/www/
 
--- a/component_down.py
+++ b/component_down.py
@@ -1,0 +1,33 @@
+import os
+import time
+import json
+from collections import defaultdict
+
+agent= os.getenv('HOSTNAME')
+now = time.mktime(time.gmtime())
+since = 7 ## in days
+
+base_dir = '/data/srv/wmagent/current/install/wmagent/'
+restarts = { 
+    "timestamp" : now,
+    "since" : since,
+    "agent" : os.getenv('HOSTNAME'),
+    "data" : defaultdict(list)
+    }
+   
+for component in os.listdir(base_dir):
+    if os.path.isfile(component):continue
+    print component
+    #for grep in os.popen('grep Harness %s/%s/ComponentLog | grep Starting | grep %s'%( base_dir, component,component)):
+    for grep in os.popen('grep terminated %s/%s/ComponentLog | grep %s'%( base_dir, component,component)):
+        timestamp = grep.split('INFO')[0][:-1].split(',')[0]
+        #2016-08-17 16:10:26,753:140679060952832
+        restart_date = time.strptime(timestamp, "%Y-%m-%d %H:%M:%S")
+        restart_time = time.mktime(restart_date)
+        if (now - restart_time) > (since*24*60*60): continue
+        print timestamp,time.asctime( restart_date )
+        restarts['data'][component].append( restart_time )
+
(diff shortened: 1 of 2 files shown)


component_down.py
```diff
@@ -0,0 +1,33 @@
...
+base_dir = '/data/srv/wmagent/current/install/wmagent/'
+restarts = { 
+    "timestamp" : now,
+    "since" : since,
+    "agent" : os.getenv('HOSTNAME'),
+    "data" : defaultdict(list)
+    }
+   
+for component in os.listdir(base_dir):
+    if os.path.isfile(component):continue
```
amaltaro: How about replacing L19 by:
```
if not os.path.exists(os.path.join(base_dir, component, 'ComponentLog'):
    continue
```

component_down.py
```diff
@@ -0,0 +1,33 @@
+import os
+import time
+import json
+from collections import defaultdict
+
+agent= os.getenv('HOSTNAME')
```
amaltaro: I think we should check whether the hostname belongs to `fnal.gov` domain or `cern.ch` and then we set the jsonDir according to it. Something like
```
if 'fnal.gov' in agent:
  jsonDir = '/afs/cern.ch/user/c/cmst1/www/Automatic_Agent_json/'
else:
  jsonDir = '/data/srv/wmagent/current'
```

such that it doesn't fail if running at FNAL nodes.

vlimant:
let me know when you need this to be merged

component_down.py
```diff
@@ -0,0 +1,35 @@
...
+    "since" : since,
+    "agent" : os.getenv('HOSTNAME'),
+    "data" : defaultdict(list)
+    }
+   
+for component in os.listdir(base_dir):
+    if os.path.isfile(component):continue
+    print component
+    #for grep in os.popen('grep Harness %s/%s/ComponentLog | grep Starting | grep %s'%( base_dir, component,component)):
+    for grep in os.popen('grep terminated %s/%s/ComponentLog | grep %s'%( base_dir, component,component)):
```
amaltaro: please check whether this file exists first.

component_down.py
```diff
@@ -0,0 +1,35 @@
...
+        timestamp = grep.split('INFO')[0][:-1].split(',')[0]
+        #2016-08-17 16:10:26,753:140679060952832
+        restart_date = time.strptime(timestamp, "%Y-%m-%d %H:%M:%S")
+        restart_time = time.mktime(restart_date)
+        if (now - restart_time) > (since*24*60*60): continue
+        print timestamp,time.asctime( restart_date )
+        restarts['data'][component].append( restart_time )
+
+open('/data/srv/wmagent/current/bin/%s.restart.json'% agent,'w').write( json.dumps( restarts, indent=2) )
```
amaltaro: It's better if you write it somewhere else, for instance `/data/srv/wmagent/logs`

vlimant:
can we have a description on how things are running on the agents ?
are we already using the code installed by puppet on agents ?
there's absolutely no issue running this under unified, we should just make it use proper directories and permissions.

mcremone:
Neither the scripts that generates the jsons nor the script that does the plot are currently running. We would need only the latter to possibly run under unified.

scarletnorberg:
The plan is to put the component_down script on ever agent. This script create a json file that is than copied to an eos area at cern. This folder is than copied on unified by the wmagent_status.py and than this script runs over the json file and makes plots.
We are not already using this code.
I agree with this: we should just make it use proper directories and permissions. yes.

component_down.py
scarletnorberg: What file? I ran it on both fermi and cern agent last night that part of the script works on both.

component_down.py
scarletnorberg: I can change that. It was more random to get the xrdcp. I will do that now do you want it in logs or some where else?

component_down.py
scarletnorberg: I do not see an issue with this will make the change run a test and re push.

component_down.py
scarletnorberg: Again. I think this part works fine it was only the xrdcp command that failed at cern but if you would like extra protections I can added these lines in and cross check to make sure it works.

scarletnorberg:
@amaltaro Did I answer all your question is using the log folder fine?

wmagent_status.py
```diff
@@ -0,0 +1,106 @@
...
+import time
+import matplotlib.cm as cm
+
+from collections import defaultdict 
+from operator import itemgetter
+import matplotlib.ticker as ticker
+
+import os
+
+os.system("xrdcp -r root://eoscms.cern.ch//eos/cms/store/logs/prod/recent/agent_crashes .")
```
vlimant: after merging this, I will change the path accordingly, the unified nodes have eos mount, so that we can just glog.glob it

scarletnorberg:
Dave and Farrukh agree that xrdcp commands should be load so are working on this. Once I run a quick test later today to make sure that it does work and nothing else needs to be changed than this should be good to go with everyone else approval of course.

PR #263, opened 2018-02-26
hook to hold sites

vlimant:
Somehow along the lines of what you suggested @bbockelm 
The agents are already removing the draining sites from site whitelist.
The purpose here is a dramatic increase in number of failures is observed for a give site to go and quickly edit it out from the whitelist, with the possibility of adding it back just as quickly.
FYI @thongonary @mcremone

### DIFF
--- a/Unified/go_condor.py
+++ b/Unified/go_condor.py
@@ -10,6 +10,8 @@ import hashlib
 import htcondor
 from collections import defaultdict
 
+
+_site_re = re.compile("(T\d)_([A-Z]{2})_([A-Z]{1}[A-Z,a-z,_]+)$")
 def makeHighPrioAds(config):
     to_be_raised = config.get('highprio',[])
     if to_be_raised:
@@ -32,8 +34,10 @@ def makeHoldSiteAds(config):
     """
     Create a rule to hold jobs from matching a given site
     """
-    held_site = config.get('hold_site',['T0_CH_CERN'])
+    held_site = config.get('hold_site',[])
+
     for site in held_site:
+        if not _site_re.match( site ): continue
         anAd = classad.ClassAd()
         anAd["GridResource"] = "condor localhost localhost"
         anAd["TargetUniverse"] = 5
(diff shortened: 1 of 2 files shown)


amaltaro:
Doesn't the P&R team has privileges to either set a site to enable or disable/drain in SSB?
Why not to just use that and let the agent make the necessary htcondor changes?

vlimant:
@amaltaro the held, as for the workflow holding (https://github.com/dmwm/WMCore/issues/8378) is a transition state to prevent automatically damage to be done while someone looks into to it.
setting drain is too invasive if there was a simple fix and the site can be released.

Unified/go_condor.py
```diff
@@ -27,6 +27,44 @@ def makeHighPrioAds(config):
...
+    """
+    Create a rule to hold jobs from matching a given site
+    """
+    held_site = config.get('hold_site',['T0_CH_CERN'])
+    for site in held_site:
+        anAd = classad.ClassAd()
+        anAd["GridResource"] = "condor localhost localhost"
+        anAd["TargetUniverse"] = 5
+        anAd["Name"] = str("Holding jobs from %s"%site)
+        anAd["Requirements"] = classad.ExprTree('regexp("%s",DESIRED_Sites) && HasBeenHeldFrom%s isnt true'% (site, site))
```
bbockelm: Can we put `site` through a regex to make sure `HasBeenHeldFrom%s` is valid?

Unified/go_condor.py
```diff
@@ -27,6 +27,44 @@ def makeHighPrioAds(config):
...
+    Create a rule to add a site back in site whitelist
+    """
+    relase_site = config.get('release_site',['T0_CH_CERN'])
+    for site in relase_site:
+        anAd = classad.ClassAd()
+        anAd["GridResource"] = "condor localhost localhost"
+        anAd["TargetUniverse"] = 5
+        anAd["Name"] = str("Releasing jobs for %s"%site)
+        anAd["Requirements"] = classad.ExprTree('HasBeenHeldFrom%s is true'% site )
+        anAd["copy_DESIRED_Sites"] = "Editing_DESIRED_Sites"
```
bbockelm: Can we call this `Releasing_DESIRED_Sites` instead?

vlimant:
it's fine to have the site put in the hold list then in the release list.

vlimant:
I thought it was fine, but now indeed I see the impracticality, not of having to put it in the "release site list", since it is not clear when to remove from the release site list, a site that was held-then-released.

Issue #277, opened 2018-05-14
Check the file size when splitting


thongonary:
Please add another condition to the splitting: the file size must be less than 20 GB.

PR #279, opened 2018-05-16
Restrict output size to a single size

vlimant:
@thongonary you'll have to check whether it fits #277 
@SvenjaPflitsch N.B.  GB_space_limit unified configuration will have to be edited down to 20

### DIFF
--- a/utils.py
+++ b/utils.py
@@ -6254,8 +6254,8 @@ class workflowInfo:
         hold = False
         ## for those that are modified, add it and return it
         modified_splits = []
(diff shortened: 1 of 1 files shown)


vlimant:
@thongonary you wanted this, do you still want it ?

thongonary:
Paus wanted this and I think he still does, so yes.

thongonary:
Yeah that looks right, except that I don't understand why checkWorkflowSplitting() gets removed.

PR #282, opened 2018-06-06
manual clone

vlimant:
@prozober to solve the issue with manual cloning. To be tested

PR #283, opened 2018-06-07
Fix all cloning

vlimant:
@prozober to be tried to fix what you've seen in testing #282 
@areinsvo @SvenjaPflitsch there is some level of testing that needs to be done in actor.
rejector.py should be untouched since I used its code centrally.


vlimant:
@prozober let one know here next time there is a simple clone to be made in actor or by hand so that we can test the other two

PR #286, opened 2018-07-09
Adding T0_CH_CSCS_HPC to overflow-sites

SvenjaPflitsch:


SvenjaPflitsch:
The @mcremone @thongonary @vlimant  I have added the mapping for CSCS. There were some mappings that were done for T0 but not HLT, which I have listed below. Pleas let me know if they should also be added. 

Line 99-100 of org-file: 
---------------------
## temptatively
mapping['T0_CH_CERN'].append( 'T2_CH_CERN' )

Line 118 of org file: 
--------------------
mapping['T0_CH_CERN'].append( 'T1_US_FNAL' )

thongonary:
Thanks Svenja. Everything looks fine to me. I don't think having CSCS overflows to T2 CERN and FNAL is necessary, but I could be wrong. If there is no objection from Jean-Roch and Matteo by the end of today, I will merge this PR.

vlimant:
https://cms-gwmsmon.cern.ch/prodview/T2_CH_CSCS_HPC is already in production
https://cms-gwmsmon.cern.ch/prodview/T0_CH_CSCS_HPC on the other hand is the site that needs to be used asap.

thongonary:
Oh thanks for spotting that! @SvenjaPflitsch can you please change T2 to T0?

SvenjaPflitsch:
Sure I will do it right away.

mcremone:
And please no overflow from CSCS to T2 CERN and FNAL

Unified/equalizor.py
```diff
@@ -40,6 +40,10 @@ def equalizor(url , specific = None, options=None):
     if options.hlt: use_HLT = True
     if use_HLT: over_rides.append('T2_CH_CERN_HLT')
     
+    use_CSCS = ('T2_CH_CSCS_HPC' in UC.get("site_for_overflow"))
+    if options.cscs: use_CSCS = True
+    if use_CSCS: over_rides.append('T2_CH_CSCS_HPC')
```
thongonary: This also needs to be changed

Unified/equalizor.py
```diff
@@ -40,6 +40,10 @@ def equalizor(url , specific = None, options=None):
     if options.hlt: use_HLT = True
     if use_HLT: over_rides.append('T2_CH_CERN_HLT')
     
+    use_CSCS = ('T2_CH_CSCS_HPC' in UC.get("site_for_overflow"))
```
thongonary: and here of course

Issue #309, opened 2018-09-14
Workflows stuck in away (Unified) / completed (ReqMgr2)


thongonary:
At the moment, there are 1589 workflows stuck in this state
https://dmytro.web.cern.ch/dmytro/cmsprodmon/requests.php?in_production=1&rsort=2&older=2&status=validation 
Some has been in this state for more than 3 weeks, for example: https://dmytro.web.cern.ch/dmytro/cmsprodmon/workflows.php?prep_id=ReReco-Run2016B-L1MinimumBias1-07Aug17_ver2-0004
Checkor runs normally -- it ran yesterday without any crash. 
We need to investigate and fix this problem.
@SvenjaPflitsch @vlimant @mcremone @dabercro

SvenjaPflitsch:
I tried to run checkor manually on the example given by Thong and got an error. 
I was logged in at vocms0275. It does not look like a general issue as the normal checkor logs do not show these kinds of error messages. @vlimant do I need to load another module or initialize something to run checkor manually now? 

-------------------------------------------------------------------------------------------
python Unified/checkor.py prebello_Run2016B-v2-L1MinimumBias1-07Aug17_ver2_8029_180504_205941_1258
Using the rw account
no options passed, assuming we do everything
Traceback (most recent call last):
  File "Unified/checkor.py", line 1338, in <module>
    checkor(url, spec, options=options)
  File "Unified/checkor.py", line 34, in checkor
    fDB = closeoutInfo()
  File "/data/unified/WmAgentScripts/Unified/utils.py", line 2517, in __init__
    import pymongo,ssl
ImportError: No module named pymongo
------------------------------------------------------------------------------------

thongonary:
@SvenjaPflitsch would these help? https://github.com/CMSCompOps/WmAgentScripts/blob/master/cWrap.sh#L49-L52

SvenjaPflitsch:
Yes that helped. Thank you Thong.

SvenjaPflitsch:
It looks like the explanation for this is rather simple in the end. 
Checkor is only handling 200 workflows per round (value is set in unified) and it is handling the ones with high priority first. And since checkor is most affected from the eos instabilities, and we have lots of high-priority work in the system it is not getting to the ones with 85k priority or less.

thongonary:
[...]

PR #317, opened 2018-09-25
threaded checkor

vlimant:
is this going to help on https://github.com/CMSCompOps/WmAgentScripts/issues/309

### DIFF
--- a/Unified/checkor.py
+++ b/Unified/checkor.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 from assignSession import *
-from utils import getWorkflows, workflowInfo, getDatasetEventsAndLumis, findCustodialLocation, getDatasetEventsPerLumi, siteInfo, getDatasetPresence, campaignInfo, getWorkflowById, forceComplete, makeReplicaRequest, getDatasetSize, getDatasetFiles, sendLog, reqmgr_url, dbs_url, dbs_url_writer, getForceCompletes, display_time, checkMemory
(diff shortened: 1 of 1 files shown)


thongonary:
I'm going to merge this by the end of today if there is no objection.

Unified/checkor.py
```diff
@@ -1326,6 +1466,7 @@ def upward( ns ):
     parser.add_option('--tape_size_limit', help='The limit in size of all outputs',default=0,type=int)
     parser.add_option('--html',help='make the monitor page',action='store_true', default=False)
     parser.add_option('--no_report',help='Prevent from making the error report',action='store_true', default=False)
+    parser.add_option('--threads',help='Number of threads for processing workflows',default=1, type=int)
```
thongonary: Is it ok if we set the default to be 5?

vlimant: I have not done any scale or scaling test of this, so any number would do, once robustly tested

thongonary: @SvenjaPflitsch can you please test this on 1 workflow to see if it runs OK with multi threads?

vlimant: you obviously have to try with more than 1 workflow otherwise there is nothing to put on multiple threads ...

thongonary:
I will merge this to test for a day. If something breaks the next day I'll rollback the PR.

SvenjaPflitsch:
I have checked the latest checkor logs and it does not look like there are any big issues. The module is running smoothly at the moment.  12h is a bit early to draw definitive conclusions but for certain options like "review-recovering" things are about 100s faster per workflow. I will keep an eye on things and check the average running time for the different options before and after this pull-request was merged.

PR #336, opened 2018-11-07
Campaign batch mongodb

vlimant:
vlimant@vocms0275.cern.ch >python campaignsConfiguration.py --help
Usage: campaignsConfiguration.py [options]

Options:
  -h, --help            show this help message and exit
  --dump=DUMP           dump the whole content in this file
  --load=LOAD           synchronize the db with the content of this file
  -n NAME, --name=NAME  a campaign name to be viewed/added/updated
  --remove              remove the specified campaign
  -c CONFIGURATION, --configuration=CONFIGURATION
                        a json doc for the campaign to be add/updated
  -p PARAMETER, --parameter=PARAMETER
                        a single parameter to be updated of the form key:value
                        or a.b.key:value for nested
can be used to update the campaign configuration, from a full json, or a single json for a new campaign, or updating one parameter only

### DIFF
--- a/campaignsConfiguration.py
+++ b/campaignsConfiguration.py
@@ -34,8 +34,12 @@ if options.load:
 if options.dump:
     uc = {}
     for content in db.find():
-        content.pop("_id")
+        i=content.pop("_id")
         if content.get('type',None): continue ## no relval
+        if 'name' not in content:
+            db.delete_one({'_id': i})
+            print "dropping",i,content,"because it is malformated"
+            continue
         uc[content.pop("name")] = content
 
     open(options.dump,'w').write(json.dumps( uc, indent =2))
@@ -51,7 +55,7 @@ if options.remove:
 post = {}
 if options.configuration:
     post.update(json.loads(options.configuration))
-
+    post['name'] = options.name
 update = {}
 if options.parameter:
     name,value = options.parameter.split(':',1)
--- a/campaignsConfiguration.py
+++ b/campaignsConfiguration.py
@@ -7,10 +7,10 @@ import sys
 parser = optparse.OptionParser()
 parser.add_option('--dump',help="dump the whole content in this file",default=None)
 parser.add_option('--load',help="synchronize the db with the content of this file", default=None)
-parser.add_option('-n','--name')
+parser.add_option('-n','--name', help='a campaign name to be viewed/added/updated')
 parser.add_option('--remove', help="remove the specified campaign", default=False, action='store_true')
-parser.add_option('-c','--configuration')
-parser.add_option('-p','--parameter')
+parser.add_option('-c','--configuration', help='a json doc for the campaign to be add/updated')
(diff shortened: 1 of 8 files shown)


vlimant:
vlimant@vocms0275.cern.ch >python campaignsConfiguration.py --dump campaigns.json

may be used to update the json that is put in github, in case people deem this necessary

vlimant:
this removes another string to eos @SvenjaPflitsch if you have time to go and do the test, we can push this in

vlimant:
@SvenjaPflitsch let's do this, so that we can put it behind

vlimant:
when you are sufficiently confident that editing campaign configuration is ok, let's move this in

SvenjaPflitsch:
I tested the different options and documented the different examples in the UO twiki under
"Using campaignsConfiguration.py "

Unfortunately at the end of the tests the "dump" option did not work any more, even though I did not change anything in the code that should affect it and it worked in the beginning. 

python campaignsConfiguration.py --dump campaignTest6.json
  File "campaignsConfiguration.py", line 39, in <module>
    uc[content.pop("name")] = content
KeyError: 'name'

campaignsConfiguration.py
```diff
@@ -50,7 +50,7 @@
 
 post = {}
 if options.configuration:
-    post.update(json.loads(options.value))
+    post.update(json.loads(options.configuration))
```
vlimant: you would need to add post['name'] = options.name

vlimant:
@SvenjaPflitsch I think https://github.com/CMSCompOps/WmAgentScripts/pull/336/commits/4673a50520bb433f0f7cb93a52b36db441255e9b#r236213838 is the reason.
I would --load campaigns.json and see whether it dumps back properly

SvenjaPflitsch:
I tried --load and afterwards --dump, but the error remains the same.

SvenjaPflitsch:
Thanks, it is working now. I will redo some of the tests just to be sure but then it should be ok. I will need something like 20 min.

SvenjaPflitsch:
Ok everything works without any errors.

vlimant:
can you load the relval in mongodb and I merge

SvenjaPflitsch:
RelVal campaigns have been added.

vlimant:
let's see how things go. you might want put the output of --dump in campaigns.json and push it to github for safety

SvenjaPflitsch:
campaigns.json is up to date. I made a copy and a --dump version to be sure.

SvenjaPflitsch:
It looks like the Autumn18 campaigns were not uploaded to the database using the --load option. I re-checked the campaigns.json file and they are definitely in the one I used for the --load option yesterday.
I tried it again to be sure, but the result is the same. 

However, when I add them one-by-one with the --configuration option, it seems to be alright. They show up with the --name option a well as with the --dump option.

PR #337, opened 2018-11-15
people should read status from DBs

vlimant:
removing a couple of files on eos, in favor or direct access in oracle of mongodb

### DIFF
--- a/Unified/htmlor.py
+++ b/Unified/htmlor.py
@@ -6,7 +6,7 @@ import os
 import json
 from collections import defaultdict
 import sys
-from utils import monitor_dir, base_dir, phedex_url, reqmgr_url, monitor_pub_dir, unified_url_eos, monitor_eos_dir, monitor_pub_eos_dir, base_eos_dir, closeoutInfo, agent_speed_draining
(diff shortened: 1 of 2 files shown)


vlimant:
@dabercro is the console using statuses.json or oracle ?

dabercro:
The old console is still using statuses.json. The new one isn't yet, but we will need a way to filter out old workflows.

dabercro:
@vlimant Can you change this back? It was probably a bad idea to make major changes while I was on holiday. Now Sharad is telling me that the workflows are backing up again. If you really want to trim this stuff, we'll either need the new console first or a larger effort on the old one.

vlimant:
@dabercro I thought the console was reading from the db directly.
the file can be added back ; can you point to where the file is accessed and we try to hack in the access from the read-only account

vlimant:
the new console should use the direct access BTW and not even try to resort to that json file

drkovalskyi:
I provided an example how to get this information easily from the database. So Dan should be able to just patch the code.

dabercro:
Yeah, I pointed to that in https://github.com/CMSCompOps/WorkflowWebTools/issues/116 At a first glance, it didn't catch right out of the box, but I didn't have time to figure out why. Hopefully I have time today.

dabercro:
It looks like everything is reading fine from the database now. Tomorrow I can confirm with Sharad that everything's working at our morning meeting.

vlimant:
please continue on https://github.com/CMSCompOps/WorkflowWebTools/issues/116

PR #338, opened 2018-11-15
mitigate file injection delay from the agent

vlimant:
related to https://github.com/dmwm/WMCore/issues/8148 and the issue with not knowing when to deal with a file mismatch, there's a 2 days grace period added in unified before labelling "filemismatch". It will label "agentfilemistmatch" in the meantime


vlimant:
@sharad1126 FYI

sharad1126:
Noted. Thanks for the information.

vlimant:
@sharad1126 there is something wrong with the labeling for now #340

PR #347, opened 2019-01-10
charmonium rereco at cern

vlimant:
@drkovalskyi  once the require Charmonium blocks are at CERN, this can be merged.
that serves as a recent example to add an adhoc jobrouting rule via unified @sharad1126


vlimant:
that's one way of draining all jobs from for these workflows from swierk to cern

sharad1126:
Transfer to CERN have been completed. Thanks @vlimant for the updates.

PR #349, opened 2019-01-11
adding muoneg to send it to T2_CH_CERN

sharad1126:
@thongonary can you have a look. I have added MuonEG to the file to make it run at CERN.

@vlimant FYI


thongonary:
Looks good to me. I think we need to make a subscription of the input datasets to CERN before merging.
/MuonEG/Run2018B-v1/RAW
/MuonEG/Run2018C-v1/RAW
Can you please do it @sharad1126

sharad1126:
data has been transferred. Merging the request.

PR #355, opened 2019-01-17
timeout on unified process

vlimant:
with a default of 24h


vlimant:
works only is ran with a user in pwdless sudo

PR #358, opened 2019-01-29
Trying to exit threads

vlimant:
@sharad1126 take a look. sounds sane to me so far


sharad1126:
looks good to me too. should we merge it?

Issue #360, opened 2019-01-31
bypass, force-complete and hold interface


vlimant:
currently from file in public afs, this should be done through a mongodb collection, which can be cleaned automatically when a workflow is set to closed-out

Issue #273, opened 2018-04-10
Reverting on-hold action


prozober:
If an "on-hold" action is taken over a workflow, we need to be able of put it back to "manual-assistance", in order to take another action. @dabercro @areinsvo @vlimant

vlimant:
please don't use the functionality from the wtc-console until we solve this

sharad1126:
I think we should merge it.

PR #364, opened 2019-01-31
HBF interface

vlimant:
goes in the direction of #360 . @sharad1126 you should get used to the command line tool.
there are some cleanup to be finalized so that we delete elements from the db when the workflow(s) go out of scope

### DIFF
--- a/Unified/actor.py
+++ b/Unified/actor.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 from assignSession import *
-from utils import workflowInfo, sendEmail, componentInfo, campaignInfo, unifiedConfiguration, siteInfo, sendLog, setDatasetStatus, moduleLock, invalidate
+from utils import workflowInfo, sendEmail, componentInfo, campaignInfo, unifiedConfiguration, siteInfo, sendLog, setDatasetStatus, moduleLock, invalidate, wtcInfo
 from utils import closeoutInfo, userLock
 import reqMgrClient
 import wtcClient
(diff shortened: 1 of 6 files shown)


sharad1126:
ok you mean to bypass/force-complete/on-hold through the command line tool directly instead of adding it in the json files?

vlimant:
with this PR, the afs files are totally deprecated and one adds something with

./wtcActions.py --action hold --keyword <workflow name, or keyword>

using --pop option to remove it from the DB.
at the moment, the db is fully synchronize with whether was in the json files on afs from the allowed users.

sharad1126:
I think we can do this. It is better in all ways. Also I guess for #273 you mean we just pop it out right?

sharad1126:
@vlimant what is the keyword here? I mean it should be name of the original workflow only right?

vlimant:
either the workflow name of some text that is found in workflow names.
this could be evolved to regexp, if one feels it should ; I don't speak regexp :-(

sharad1126:
ok great.. thanks ... I'll try this out once I have some request we need to bypass/force-complete/keep on hold. I will also add this to the twiki so it would be helpful

sharad1126:
@thongonary FYI

sharad1126:
So just to update for future reference - we need the proxy and export PYTHONPATH=$PYTHONPATH:/usr/lib64/python2.7/site-packages as it is a unified script.

sharad1126:
@vlimant can you please confirm that since we are using the wtcActions.py and soon the JIRA search(once tested and merged), we don't need https://github.com/CMSCompOps/WmAgentScripts/blob/master/utils.py#L6563 and https://github.com/CMSCompOps/WmAgentScripts/blob/master/utils.py#L6497 and can be commented from the code?

vlimant:
the file interface is completely unused already anyways, so those can go away

todor-ivanov:
Thanks @vlimant @sharad1126

PR #365, opened 2019-02-01
Creating JIRA for agent error

vlimant:
this is the alpha version. there needs to be some veto on setting a JIRA.
Heartbeat ? find a time over which to make the JIRA maybe ?
Components in auto-restart ? certainly not. Which are they ?

@scarletnorberg

### DIFF
--- a/JIRAClient.py
+++ b/JIRAClient.py
@@ -69,7 +69,7 @@ class JIRAClient:
                     set_to = prios[p]
             priority = set_to
 
-        elif not priority.isdigit():
+        elif (priority and not priority.isdigit()):
             priority = { 'decision' : '6',
                          'blocker' : '1'}.get(priority, None)
         if priority:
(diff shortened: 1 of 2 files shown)


vlimant:
works : https://its.cern.ch/jira/browse/CMSCOMPPR-4607

vlimant:
it can be merged if @scarletnorberg is happy with this

amaltaro:
Isn't it better if it sets the alarm timestamp in the jira title? "week 4" doesn't look clear at all.

vlimant:
is the timestamp "unique" in this case. "week 4" allows to have only one jira per component per week per error type per agent.

amaltaro:
I see, "week 4" would make it easier to recycle tickets. Yes, timestamp in wmstats is unique.

vlimant:
I wasn't clear : the timestamp identifies the crash, or when the error is reported to wmstats by the agent ?

PR #366, opened 2019-02-01
actor updating comments in JIRA

vlimant:
this will update the comments of the jira with the workflow prepid in the summary ; one comment per task might be a lot. we'll see with experience @sharad1126

### DIFF
--- a/JIRAClient.py
+++ b/JIRAClient.py
@@ -104,6 +104,7 @@ class JIRAClient:
               }.get( status , None)
         if to:
             try:
+                print jid,"to",status
                 self.client.transition_issue( jid, to)
             except Exception as e:
                 print "transition to",status,"not successful"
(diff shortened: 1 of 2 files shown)


vlimant:
with the try/catch, this can be merged anytime @sharad1126

vlimant:
something like 
ACDC created for task DataProcessing. Actions taken
{"cores": "", "sites": ["T1_UK_RAL", "T1_US_FNAL", "T2_FR_IPHC", "T2_US_Caltech", "T2_US_UCSD"], "memory": "20000"}

vlimant:
let's keep an eye on https://cms-unified.web.cern.ch/cms-unified//logs/actor/last.log for when the next action will be set @sharad1126 and see how comments are edited on jira

PR #381, opened 2019-02-11
check on jira

vlimant:
From the current outing of JIRA, modules are failing.
 should it be a soft or hard check is the question @sharad1126 @drkovalskyi

### DIFF
--- a/Unified/addHoc.py
+++ b/Unified/addHoc.py
@@ -42,7 +42,7 @@ if N_for_cloud:
 #os.system('Unified/assignor.py --from_status staging RunIISummer16DR80Premix')
 #os.system('Unified/assignor.py --from_status staging RunIISummer16DR80-')
 
-up = componentInfo(soft=['mcm','wtc'])                                 
+up = componentInfo(soft=['mcm','wtc','jira'])
 if not up.check(): sys.exit(0)
 
 url = reqmgr_url
--- a/Unified/checkor.py
+++ b/Unified/checkor.py
@@ -115,7 +115,7 @@ def checkor(url, spec=None, options=None):
     SI = siteInfo()
     CI = campaignInfo()
     mcm = McMClient(dev=False) if use_mcm else None
-    JC = JIRAClient() if up.status.get('jira',True) else None
+    JC = JIRAClient() if up.status.get('jira',False) else None
 
     ## retrieve bypass and onhold configuration
     bypasses = []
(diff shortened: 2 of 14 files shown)


sharad1126:
I think it should be a hard check. Else, if we have jira down again in future, it might again give errors and not create the tickets.

drkovalskyi:
I think it should be a soft check. Unfortunately we cannot rely on JIRA as a critical service, since it's not supported as such. The tools should be able to fail graceful when JIRA is down and when it's back up it should be able to resume.

vlimant:
hard check : it will fail gracefully when jira is down, but will do nothing useful until it's up
soft check : it will do what it is meant to do, and skip any jira manipulation.

soft or hard ?

drkovalskyi:
I didn't mean to fail the component, i.e. the component should work even if JIRA is down. So for me it's soft check. @sharad1126 do you agree? If not, can you please clarify what's wrong with my view?

sharad1126:
ok let's go with a soft check but what does 'skip any jira manipulation' exactly means?

vlimant:
no jira comment, or opening ; which indeed might be impacting @sharad1126 smooth operation.

for closing and agent-JIRA, one needs to add it to the PR

sharad1126:
yeah that's what I thought of and would have to do manual work in that case. So I'll switch back and prefer hard check

drkovalskyi:
Why JIRA comments cannot be generated when JIRA is up in the next cycle? There won't be smooth operation if the component won't run and if JIRA is down for a long time we will mess production.

vlimant:
the comments and action taken depend on the status the wf is in, at the next cycle the wf will have left the status it is in. I'd suggest a hard check (also because its the first time ever it was down) and switch to soft on the day it's down for a week and blocking ops

drkovalskyi:
I like your suggestion, but I also have an alternative one. What if we skip workflows that need action, but JIRA is down? This way it will prevent JIRA taking down critical components. Is it hard to implement?

vlimant:
too much refactoring needed act only on the things that'll need a jira action

drkovalskyi:
Ok. In this case I agree with the hard check option.

vlimant:
@sharad1126 can you please merge and keep an eye on all these modules with @scarletnorberg (UO)

sharad1126:
sure

Issue #374, opened 2019-02-07
put unified report information in mongodb


vlimant:
from showError, some document should be created to keep all information centralized and consumable downstream (recoverord, console, ...)

vlimant:
also does #376

PR #377, opened 2019-02-07
towards getting the new console in operation

vlimant:
a first step for #339

PR #383, opened 2019-02-15
Report in mongo

vlimant:
towards #374


vlimant:
ready once #377 is done to set action via rules.
there might be the need for changing manual => acted on the unified status.
https://github.com/CMSCompOps/wtc-console/issues/17 will also have to be done so that one can see that actions are already being in place, and avoid duplicates

vlimant:
https://cms-unified.web.cern.ch/cms-unified/report/pdmvserv_task_HIG-RunIIFall18wmLHEGS-01126__v1_T_190128_085854_2388
with a couple of goodies

Issue #384, opened 2019-02-19
JIRA from completor


vlimant:
one can set a priority and time threshold over which a JIRA is created automatically

vlimant:
@drkovalskyi if you have suggestions on changing the values for the notification threshold

PR #385, opened 2019-02-19
Create JIRA for long running workflows

vlimant:
for #384 . With  
```

jira_priority_and_delays = { 11000 : 14,
                                 90000 : 21}
```
it is going to make a jira for wf with prior>=110k after 14 days, >= 90 after 21days, and nothing else.
With
`         ping_on_jira = 7 *(24*60*60) # 7 days `
it will insert a new comment every 7 days after the last comment in the jira

### DIFF
--- a/JIRAClient.py
+++ b/JIRAClient.py
@@ -25,7 +25,34 @@ class JIRAClient:
             print "That ain't going to work out"
             sys.exit(1)
         self.client = jira.JIRA('https://its.cern.ch/jira' , options = {'cookies':  cookies})
-    
+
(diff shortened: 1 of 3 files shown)


vlimant:
https://its.cern.ch/jira/browse/CMSCOMPPR-4808 for a test

vlimant:
the calculation of the delay in this case should be taken from the last time the priority was set the current value

vlimant:
@drkovalskyi do you want this in ?

drkovalskyi:
As we discussed offline the answer was yes :)

PR #387, opened 2019-02-19
Closing blocks when requested for processing

vlimant:
instead of having opened blocks in phedex dangling around

### DIFF
--- a/Unified/assignor.py
+++ b/Unified/assignor.py
@@ -292,6 +292,8 @@ def assignor(url ,specific = None, talk=True, options=None):
                 print "endpoints from stagor",dataset_endpoints[prim]
                 endpoints.update( dataset_endpoints[prim] )
             set_lfn = getLFNbase( prim )
+            ## if they are requested for processing, they should bbe all closed already
+            closeAllBlocks(url, d, blocks)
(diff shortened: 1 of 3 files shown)


vlimant:
@amaltaro tell me if you see any drawbacks to this

amaltaro:
Yes and it should not be forced this way. DBS injection depends on the PhEDEx blocks to be closed, PhEDExInjector is not checking if a block is closed or not, but it's actually closing blocks after some requirements are matched. In short, this won't speed up anything.

The correct way to get this done is to fix that issue in WMCore
https://github.com/dmwm/WMCore/issues/8148

vlimant:
yes to making this more resilient in wmcore.
merging since there is no otherwise drawbacks

amaltaro:
Are you sure that a block not completed (without a complete replica somewhere) is the same as a block being still open?

amaltaro:
BTW, if you want to avoid issues with this. It's better to check whether those blocks opened in PhEDEx are also present (and closed in DBS). If they are not, then you might be closing things before it's supposed to.

Another question, do you record which blocks go through this Unified update? Might be useful for any future debugging.

vlimant:
there's no record ; if people say how they want this, it can be added.

Issue #346, opened 2018-12-19
speed up duplicate lumi check


vlimant:
from @yuyiguo 

DBS provide listBlocks API (https://cms-http-group.web.cern.ch/cms-http-group/apidoc/dbs3-client/current/dbs.apis.html?highlight=listblocks#dbs.apis.dbsClient.DbsApi.listBlocks). This API allows one to list all the blocks of a dataset. Because a dataset may have a huge number of blocks that across a long time period so one may use the date information in this API to limit the interesting blocks.  This date information is especially designed for the production team to inspect certain data in a period.

After one gets the interesting blocks, then one can use DBS API listFileLumis (https://cms-http-group.web.cern.ch/cms-http-group/apidoc/dbs3-client/current/dbs.apis.html#dbs.apis.dbsClient.DbsApi.listFileLumis).  This API will list the run, lumi_section, lfn and so on. There is no needs to get the run numbers.

#loop over blocks from the dataset
For block in dataset in (t1 t2)
     #get  (lfn,lumi, run) for each block
    listFilelumis(block) 

instead of 
https://github.com/CMSCompOps/WmAgentScripts/blob/master/utils.py#L4468

PR #389, opened 2019-03-01
faster DBS loop

vlimant:
is this what you had in mind @yuyiguo #346


vlimant:
about 5 times faster on /SingleMuon/Run2018D-v1/RAW for example

@sharad1126 fyi

yuyiguo:
You might use date/time to limit the blocks and it could be even faster if you were only interested to certain time or you could add new times each time.  I don't know the details your use case.

vlimant:
we want all blocks

yuyiguo:
Got it.

Issue #297, opened 2018-08-03
Alert system for failed workflows/ACDCs


sharad1126:
WMStats has a link which lists the failed workflows/ACDCs as soon as it fails. Can we have some type of alerts that helps us to know about it so an immediate action can be taken?

Link for WMStats -> **request alert** tab(on top of [https://cmsweb.cern.ch/wmstats/index.html](url)) and when we scroll down, at the bottom, it shows all the failed workflows/ACDCs under a heading _**failed**_.

We can somehow fetch it and send alerts about it on cms-comp-ops or something similar. 

@SvenjaPflitsch @amaltaro @prozober 

Can you please have a look and suggest what can be done? Thanks.

PR #390, opened 2019-03-01
Check failed ACDC

vlimant:
@sharad1126 for #297


sharad1126:
this will be really helpful.

PR #398, opened 2019-03-22
Consider block, run and lumi list in accounting

vlimant:


vlimant:
@sharad1126 @drkovalskyi

sharad1126:
If I'm right, this is going to help us with the staging/transfer issues right

PR #408, opened 2019-04-18
Enabling stepchain conversion.

mcremone:


vlimant:
have you recently tried to to a + between a string and a dict ?

PR #409, opened 2019-04-19
Factor cycle

vlimant:
some cleanup of the cycle scripts
@drkovalskyi @sharad1126


vlimant:
is it ok to send all emails back to cmsunified@cern.ch instead to individuals @drkovalskyi ?

PR #410, opened 2019-04-26
failed workflows should have a tag to take actions

sharad1126:
When request manager has workflows with status failed, we should get it in assistance manual as tags like 'manual inconsistent' so we can take actions on it without depending on UO to check it in logs and then report. 

It does send to critical logs though. https://github.com/CMSCompOps/WmAgentScripts/blob/master/Unified/checkor.py#L599

@thongonary @vlimant can you please have a look?
@drkovalskyi  FYI


thongonary:
OK for me. Just wonder does it make sense for the `inconsistent`tag though? It seems to be designed for checking the inconsistent output statistics rather than failed workflow, but it's up to you to define the convention.

PR #411, opened 2019-04-26
Take filter efficiency into account when splitting based on size/evt

thongonary:
The output size should be post efficiency calculation


thongonary:
@sharad1126 perhaps you want to double check my logic just in case.

sharad1126:
using efficiency_factor in conditions should help I guess. looks good to me.

PR #417, opened 2019-04-29
specific unlocked assigned to anaops

vlimant:
@drkovalskyi for what concerns all dataset left over in dataops by dynamo, this will clear things up little by little and will prevent from things getting there again.
@dr-stringfellow FYI

### DIFF
--- a/Unified/remainor.py
+++ b/Unified/remainor.py
@@ -122,11 +122,14 @@ if sys.argv[1] == 'parse':
                 only_unlock.add(item[0])
         table+="</table></html>"
         eosFile('%s/remaining_%s.html'%(monitor_dir,site),'w').write( table ).close()
-        for item in only_unlock:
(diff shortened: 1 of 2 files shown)


vlimant:
essentially, everything that will have only "unlock" on the pages like https://cms-unified.web.cern.ch/cms-unified/remaining_T1_DE_KIT_Disk.html and for specific data tiers, will have their dataops, 100% subscriptions be change to anaops ; as one would expect.

vlimant:
merging to have the extra code. all optional in mainstream with https://github.com/CMSCompOps/WmAgentScripts/pull/417/commits/c694af97cd7512463c40707b0af14425557ffb7b

vlimant:
looking at https://github.com/SmartDataProjects/dynamo/issues/256 it does not seem it will even help, since dynamo does not see the change of subscriptions owner

PR #418, opened 2019-05-01
Use single dbsApi/dbsApiWriter instance through whole code

vkuznet:


vlimant:
Is this Thread.threading safe ?

vkuznet:
It is actually question to Yuyi, as far as I can tell DbsApi should be thread safe. It setups persistent connection that can be re-used. Since in majority case you're fetching data there is no harm to re-use single instance, but for writer it is more important question.

vlimant:
@vkuznet can you please rebase ?

vkuznet:
It will take me a whole day to merge, therefore I'll delete my fork and then re-submit the changes. Here I'll close this PR, new one will appear soon.

PR #419, opened 2019-05-03
removed jira for heartbeats

sharad1126:
@thongonary please have a look? I have commented out the code that creates jiras for heartbeats. 

@amaltaro  FYI


thongonary:
Looks ok to me.

vlimant:
this is shutting down completely the making of jira for any agent issue, including the hard failures, with a component down.

sharad1126:
so should we just put the condition if alert_type!=heartbeat, then it makes a jira?

sharad1126:
Guess this check should help.

PR #424, opened 2019-05-22
Correct events per lumi based on time per event

thongonary:
Confirmation from PPD

> Imagine we want to generate 10 events (each has 5 s time/event) and the request has filter efficiency of 0.5. Sample will actually produce 20 events and the time/event has no correction (still 5s and not 5s / 0.5)

So if the target time limit is 100 sec, the `max_events_per_lumi` allowed is (time limit) / (time per event) = 100/5 = 20 events, which will give 10 events in the output lumi. (This is what the PR is proposing.)

If we allow `max_events_per_lumi` to be divided by `efficiency_factor` like it has been currently, we will let `max_events_per_lumi` to be 100/5/0.5 = 40 events, which will take 200 secs to finish, ie, double the limit.

### DIFF
--- a/utils.py
+++ b/utils.py
@@ -7630,7 +7630,7 @@ class workflowInfo:
                                                                                                                                                      time_per_input_lumi / (60.*60.),
                                                                                                                                                      job_timeout), level='critical')
                             this_max_events_per_lumi = int( job_target*60.*60. / timeperevent)
-                            max_events_per_lumi.append( this_max_events_per_lumi / efficiency_factor) # to be tested
+                            max_events_per_lumi.append( this_max_events_per_lumi ) 
 
                         else:
                             pass
--- a/utils.py
+++ b/utils.py
@@ -7613,7 +7613,9 @@ class workflowInfo:
                     #    #max_events_per_lumi.append( int(avg_events_per_job*0.75) ) ##reducing
                     #    max_events_per_lumi.append( int(avg_events_per_job*0.75) ) ##reducing
 
-
+		    # Time_per_event as input from PPD does not take into account of the filter efficiency. 
+		    # Therefore the time constraint must be number of actual events that the job needs to run on
+		    # ie, before the filter efficiency, not the number of output events (ie, after the filter efficiency).
                     if timeperevent:
                         job_timeout = 45. ## hours
                         job_target = 8. ## hours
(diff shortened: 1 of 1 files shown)


sharad1126:
is filter efficiency the same as efficiency_factor?

thongonary:
Yes

sharad1126:
the change you said would apply only
```
 if (time_per_input_lumi > (job_timeout*60*60)):
```
https://github.com/CMSCompOps/WmAgentScripts/blob/master/utils.py#L7621

sharad1126:
we are using https://github.com/CMSCompOps/WmAgentScripts/blob/master/utils.py#L7608 as the first task has eventsperlumi and eventsperjob

thongonary:
> the change you said would apply only
> 
> ```
>  if (time_per_input_lumi > (job_timeout*60*60)):
> ```
> 
> https://github.com/CMSCompOps/WmAgentScripts/blob/master/utils.py#L7621

Yes






> we are using https://github.com/CMSCompOps/WmAgentScripts/blob/master/utils.py#L7608 as the first task has eventsperlumi and eventsperjob

I think the events per lumi calculation needs to be treated differently for time constraint and size constraint. For time constraint we care about number of events **before** filter efficiency. For size constraint we care about number of events **after** filter efficiency. 

Therefore `time_per_input_lumi` in https://github.com/CMSCompOps/WmAgentScripts/blob/master/utils.py#L7620 should be calculated as `events_per_lumi` x `timeperevent` (removing `efficiency_factor` in https://github.com/CMSCompOps/WmAgentScripts/blob/master/utils.py#L7608).

While for `size_per_input_lumi` in https://github.com/CMSCompOps/WmAgentScripts/blob/master/utils.py#L7640 should be equal to `events_per_lumi` x `timeperevent` x `efficiency_factor` (ie, keep the `efficiency_factor` like it is now).

sharad1126:
Also I think we should multiply the efficiency factor for time_per_input_lumi in timeperevent constraint as it is for input. 
https://github.com/CMSCompOps/WmAgentScripts/blob/master/utils.py#L7620

thongonary:
> I think this is fine to merge as we don't want to decrease the event/Lumi here if filter efficiency is < 1.
Label "validation needed" means we should run this in a test machine before merging. 



> We should let this divided by the efficiency factor so we have more events/Lumi in this case. taking more time isn't an issue as compared to having small lumi, right?
What? Taking more time **is** an issue, even a bigger issue (job will crash if exceeding 44 hours). 


> Also I think we should multiply the efficiency factor for time_per_input_lumi in timeperevent constraint as it is for input.
> https://github.com/CMSCompOps/WmAgentScripts/blob/master/utils.py#L7620
As I explained in the beginning of this PR, time constraint should not be concerned with filter efficiency.

sharad1126:
Got your point. I will try to test these changes on the test machine.

thongonary:
Validated. The key change is to have 2 different values of filter efficiencies: one for the input tasks and one for the current task, and apply them properly. I will merge now.

sharad1126:
That's great.

PR #426, opened 2019-06-04
switch to kill and reset acronjobs for unified

sharad1126:
Fixes kill switch for unified

#### Status
ready

#### Description
use the simple script to kill or reset the acronjobs for unified

yes

#### Related PRs
NA

https://raw.githubusercontent.com/CMSCompOps/WmAgentScripts/master/acrontab.list

@thongonary @vlimant @drkovalskyi

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -9,14 +9,10 @@ Setting Up Scripts
 Download the Scripts
 ~~~~~~~~~~~~~~~~~~~~
 
-The easiest wy to download the WmAgentScripts is using git on lxplus or your own machine::
+The easiest way to download the WmAgentScripts is using git on lxplus or your own local machine::
 
     git clone https://github.com/CMSCompOps/WmAgentScripts.git
 
-.. Note::
-
-  Does this work? I know a lot of things depend on dbsClient.
-  OpsSpace tries to install this, but it's not tested yet.
 
 Creating Proxy
 ~~~~~~~~~~~~~~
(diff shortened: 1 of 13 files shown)


drkovalskyi:
I find it a bit kludgy and confusing. First I don't see any kills, i.e. you don't do what you say - you simply nuke acrontab config. It's still probably a neat useful script, but I would expect a check within Unified scripts for a global lock file not to run is a cleaner solution. In the current form any manual edits to the acrontab will be lost and you won't be able to have different acrontab configurations on different machines.

thongonary:
> In the current form any manual edits to the acrontab will be lost and you won't be able to have different acrontab configurations on different machines.

Doesn't the centralized config on github (https://github.com/CMSCompOps/WmAgentScripts/blob/master/acrontab.list) already specify which processes to run on different machines? I would expect any manual edit to be synchronized to github.

thongonary:
If we want to preserve the current config then probably go with something along the lines of
```
acrontab -l > current_config
sed 's/^/#/' current_config | acrontab
```
then to turn it back on
```
acrontab < current_config
```

sharad1126:
Hi @drkovalskyi , @thongonary ,
I made this script not to kill the running jobs but to stop the acronjobs and then reset them so when we have cmsweb upgrades, we don't get unlimited failure emails or whenever we want(if we want to stop running the acronjobs/modules), we can just pause all acronjobs. There are two options to pause them:
1. Delete all entries in acrontab 
2. Comment all entries in acrontab
Doing any of the above options will work in the similar manner i.e. deactivating further acron jobs until they are reset.
We can always reset them by either putting back the entries or removing the '#' that we inserted depending on the options we choose to pause them.

vlimant:
an existing option is https://github.com/CMSCompOps/WmAgentScripts/blob/master/cWrap.sh#L31

`touch /eos/cms/store/unified/unified_drain `

will stop anything from starting

`rm /eos/cms/store/unified/unified_drain `

will resume

vlimant:
But, the better solution is to have an effective programmatic check that cmsweb is out for maintenance in https://github.com/CMSCompOps/WmAgentScripts/blob/master/utils.py#L1179 which it is supposed to do already ;

sharad1126:
Hi @vlimant , according to what I understand(I can be wrong here), when we do the component checking the modules are still running and when a component is down, it still gives an error message which is not similar to stopping modules and sending failure messages.
For the **unified_drain** thing you mentioned, I never saw it before or had a clue that this existed (technically it just stops cWrap from running). We can document this in the GH wiki too, but what the  script in the PR is doing is pausing all acronjobs that don't use cWrap.sh too, which we call global drain.

vlimant:
hardly anything is run without cWrap

vlimant:
and component checking will prevent module from running if a critical component is not reachable

sharad1126:
can you please also tell a little more about the local drain and global drain. I will document this on the GH wiki and we can close this PR.

thongonary:
I think local drain just means all processes on that particular machine are stopped (which is why the file `unified_drain` is local on /data/unified/WmAgentScripts, ie, unique to each machine) vs. global drain where all processes on all machines are stopped (`unified_drain` on /eos -- accessible from all machines).

PR #429, opened 2019-06-14
Change console link from seeworkflow->seeworkflow2 and comment other consoles.

sharad1126:
Fixes the issue for manually adding 2 after the seeworkflow on opening the console.

#### Status
ready

#### Description
The console has two sub urls: seeworkflow and seeworkflow2. seeworkflow2 was made inorder to see that the deadlock doesn't occur and we don't get the 404 error page that we often get with seeworkflow. The other console links are never used as the new console never got ready to be used till now, so we can comment out those for now.

yes

@thongonary can you please have a look?


sharad1126:
@thongonary can you please have a look here?

thongonary:
Looks good to me -- I assume Dan already confirmed that seeworkflow2 is the one that is better maintained.

sharad1126:
yes, @dabercro made seeworkflow2 because we used to get 404 error in seeworkflow.

PR #430, opened 2019-06-25
Change json file with campaigns provided by PPD

sharad1126:
Fixes issue of too many unwanted(not needed anymore) campaigns.

#### Status
ready

#### Description
The JSON file has been changed with all the campaigns that were listed by PPD in https://docs.google.com/document/d/1NWp4rhDiZAJ4c9YCGaPVRjMQgwhfWaJvUpB8Yq7RNQ4/edit?ts=5d0b337e#
except for HINPbPbAutumn18DRHiMix that doesn't exist in our configs(no JIRA and no wfs submitted with this campaign name).

#### Is it backward compatible (if not, which system it affects?)
yes, unless it's not put into the campaigns' configuration in the MongoDB. 

#### Related PRs
none

#### External dependencies/deployment changes
the JSON file needs to be loaded into the MongoDB to make changes in the system. 

#### Mention people to look at PRs
@thongonary can you please have a look?
@optimisticcynic FYI

PR #433, opened 2019-07-05
script to remove campign and put in archive

sharad1126:
Feature to remove campaign from mongodb, and campaigns.json and copy to archive_campaigns.json

#### Status
tested and ready 

#### Description
The script helps to remove the unwanted campaign from the mongodb,campaigns.json and copy it to archive_campaigns.json.

yes

#### Related PRs
https://github.com/CMSCompOps/WmAgentScripts/pull/430

needs campaigns.json, archive_campaigns.json and campaignsConfiguration.py 

@thongonary @optimisticcynic can you please have a look?

### DIFF
--- a/remove_campaign.py
+++ b/remove_campaign.py
@@ -11,28 +11,28 @@ parser.add_option('-n','--name',help='campaign name to be removed')
 (options,args) = parser.parse_args()
 
 if not options.name:
-	print "enter a campaign name as parameter -n <campaign name>"
-	sys.exit(1)
+    print "enter a campaign name as parameter -n <campaign name>"
+    sys.exit(1)
 p=0
 with open('campaigns.json','r') as campaigns:
-	campaign_db = json.load(campaigns)
-        for key,value in campaign_db.items():
-		if key == options.name:
-                        # remove the campaign from the mongoDB
-	            	remove_from_campaign_config = 'python campaignsConfiguration.py --name ' + options.name	+ ' --remove'
-	            	os.system(remove_from_campaign_config)
-			with open('archive_campaigns.json','r') as outfile:
-				destination = json.load(outfile)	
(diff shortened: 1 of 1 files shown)


thongonary:
The code looks OK, just kinda suboptimal in terms of style.

sharad1126:
do you suggest any changes? from my side, I prefer to keep the code as efficient and small as possible so other people can understand it whenever needed.

thongonary:
I think you can open it once for both read and write ("w+")

sharad1126:
tried doing that but it gives issues with r+ or w+ in indentation and deletion. Guess this is the better way to read and write it separately when needed.

amaltaro:
My 2cents here is.
```
if not options.name:
  print "enter a campaign name with -n <campaign name>"
  sys.exit(1)

with open('campaigns.json','r') as campaigns:
  campaign_db = json.load(campaigns)
with open('archive_campaigns.json','r') as outfile:
  destination = json.load(outfile)

for key,value in campaign_db.items():
...
```

So, files being read, you open, load into memory and move on , no need to keep all that nested structure.
If you want to avoid problems in the future, do NOT use tabs in a python source code.

sharad1126:
I wouldn't like to open the archive_campaigns.json file unless I know that the campaign actually exists

amaltaro:
That's a valid point. However:
a) json.load return a dict, so there is no need to iterate over it, simply do a `if key in campaign_db` and move on
b) unneeded nested code
c) opening files inside the for loop gives the impression you would need to open it many times, but you dont
d) code can be made much cleaner and simpler

Last but not least, it's a good practice to run a code reformat on brand new source code such that you keep it tidy and following PEP coding style. Whatever editor/IDE works for you.

PR #438, opened 2019-08-05
sending emails to PPD HN for relvals < 50%

### LINKED ISSUE #436: Check on relval stat before announcement
So that relval will not announce at 2% without anyone notice.

Perhaps just modify this 
https://github.com/CMSCompOps/WmAgentScripts/blob/master/Unified/closor.py#L304
with the final statistics.

sharad1126:
Fixes #436 

#### Status
ready but nor tested

#### Description
Added a flag to check 50% and then send an extra email to HN. 

yes

no

@thongonary can you please have a look?

### DIFF
--- a/Unified/closor.py
+++ b/Unified/closor.py
@@ -299,51 +299,32 @@ def closor(url, specific=None, options=None):
         if go:
             subject = "Release Validation Samples Batch %s"% bname
             issues=""
-            if batch_warnings[ bname ]:
-                issues="The following datasets have outstanding completion (<%d%%) issues:\n\n"% batch_goodness
-                issues+="\n".join( sorted( batch_warnings[ bname ] ))
-                issues+="\n\n"
-            text = """
-Dear all,
-
-a batch of release validation workflows has finished.
-
-Batch ID:
-
-%s
-
-Detail of the workflows
-
-https://dmytro.web.cern.ch/dmytro/cmsprodmon/requests.php?campaign=%s
-
-%s 
-This is an automated message.
-"""%( bname, 
-      bname,
-      issues)
-            to = ['hn-cms-relval@cern.ch']
-            sendEmail(subject, text, destination=to )
-            ## just announced ; take it out now.
-            #BI.pop( bname )
-            subject = "Low Statistics for Relval"
-            issues = ""
+            #if batch_warnings[ bname ]:
+            #    issues="The following datasets have outstanding completion (<%d%%) issues:\n\n"% batch_goodness
+            #    issues+="\n".join( sorted( batch_warnings[ bname ] ))
+            #    issues+="\n\n"
             if batch_extreme_warnings[ bname ]:
+                subject = "Low Statistics for %s"% bname
                 issues="The following datasets have outstanding completion (<50%%) issues:\n\n"
                 issues+="\n".join( sorted( batch_extreme_warnings[ bname ] ))
                 issues+="\n\n"
-                text = ""
-                text+= "Dear all,\n\n"
-                text+= "A batch of release validation workflows has finished.\n\n"
-                text+= "Batch ID:\n\n"
-                text+= "%s\n\n"%( bname )
-                text+= "Detail of the workflows\n\n"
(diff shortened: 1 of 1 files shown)


Unified/closor.py
```diff
@@ -317,6 +319,34 @@ def rank( wfn ):
 %s 
 This is an automated message.
 """%( bname, 
+      bname,
+      issues)
+            to = ['hn-cms-relval@cern.ch']
+            sendEmail(subject, text, destination=to )
+            ## just announced ; take it out now.
+            #BI.pop( bname )
+            subject = "Release Validation low statistics"
```
thongonary: Let's put `Low statistics` first in case the email subject gets truncated on phone.

Unified/closor.py
```diff
@@ -317,6 +319,34 @@ def rank( wfn ):
...
+
+%s
+
+Detail of the workflows
+
+https://dmytro.web.cern.ch/dmytro/cmsprodmon/requests.php?campaign=%s
+
+%s 
+This is an automated message.
+"""%( bname,
```
thongonary: The second `sendEmail` is outside of `if batch_extreme_warnings[ bname ]:` condition. You will send email twice every time regardless of final statistics.

To ensure code readability, I would suggest to reformat the main body text, something like:
```
    text = ""
    text += "Dear all,\n\n" # Double linebreak since HN format usually will truncate one.
    text += "A batch of ..." 
```

And then depends on the condition (`batch_warnings` or `batch_extreme_warnings`), you will add the proper issue and change the subject accordingly, and send email only *once*.

sharad1126: @thongonary I am changing the subject line. For the email thing, I thought this has to be a second email in case it is < 50 else only one email. do you want to send only one email if the stats<50 ?

Unified/closor.py
sharad1126: Changed subject so that even it gets truncated on phone, you will see ``Low Statistics``

Unified/closor.py
```diff
@@ -322,6 +324,26 @@ def rank( wfn ):
             to = ['hn-cms-relval@cern.ch']
             sendEmail(subject, text, destination=to )
             ## just announced ; take it out now.
+            #BI.pop( bname )
+            subject = "Low Statistics for Relval"
```
thongonary: Please add the batch name in the subject line, like `"Low Statistics for %s" %(bname)`

Unified/closor.py
```diff
@@ -322,6 +324,26 @@ def rank( wfn ):
...
+                text+= "A batch of release validation workflows has finished.\n\n"
+                text+= "Batch ID:\n\n"
+                text+= "%s\n\n"%( bname )
+                text+= "Detail of the workflows\n\n"
+                text+= "https://dmytro.web.cern.ch/dmytro/cmsprodmon/requests.php?campaign=%s\n\n"%( bname )
+                text+= "%s\n\n"%(issues) 
+                text+= "This is an automated message.\n\n"
+                text+= ""
+                to = ['hn-cms-relval@cern.ch']
+                sendEmail(subject, text, destination=to )
```
thongonary: I would prefer to send 1 email only. Normal relval completion email, and in case of low stat, send one low stat email only.

thongonary: So I would remove the first email path (the unformatted block) and use `if` statement to modify some part of the subject and body, and send the email. Makes the code much leaner.

thongonary: Not really remove, but merge the 2 parts.

Unified/closor.py
sharad1126: ok sure

Unified/closor.py
sharad1126: But, I thought what we discussed in PnR meeting was to send an additional email. If you say, we change it to one email only.

sharad1126: yeah, I mean the older email format is pre-existent. I didn't touch that, just added the new one. If we want, we can change that too in a better format.

thongonary:
OK looks good.

PR #444, opened 2019-08-21
check file stat after copying

thongonary:
#### Status
<In development | not-tested >

#### Description
https://cern.service-now.com/service-portal/view-incident.do?n=INC2095005
After copying the file, should check if the destination file in EOS has file size > 0, else retry

YES 

@sharad1126

### DIFF
--- a/utils.py
+++ b/utils.py
@@ -1364,11 +1364,11 @@ class eosFile(object):
         bail_and_email = False
         T = 0
         while T < self.trials:
-	        T += 1
+            T += 1
             try:
                 print "moving",self.cache_filename,"to",self.eos_filename
                 print("Attempt {}".format(T))
-                r = os.system("cp %s %s"%( self.cache_filename, self.eos_filename))
+                r = os.system("eoscp %s %s"%( self.cache_filename, self.eos_filename))
                 if r==0 and os.path.getsize(self.eos_filename) > 0: return True
                 print "not able to copy to eos",self.eos_filename,"with code",r
                 time.sleep(30)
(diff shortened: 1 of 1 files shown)


sharad1126:
also should we use eoscp instead of cp as suggested by Maria in the ticket?

thongonary:
> are we changing the value of bail_and_email to true?
No we change it to False.

> This will still only check the size but we don't know the content. many times eos just copies '@' in the file and the file size isn't 0.
If that's true, can you report in the ticket to Maria? She claimed that when the copy fails the destination file is blank and suggested me to do stat check.

sharad1126:
> > are we changing the value of bail_and_email to true?
> 
> No we change it to False.
> 
> > This will still only check the size but we don't know the content. many times eos just copies '@' in the file and the file size isn't 0.
> 
> If that's true, can you report in the ticket to Maria? She claimed that when the copy fails the destination file is blank and suggested me to do stat check.

but then how are you going to get in the if block on line 1375?

sharad1126:
Also, as mentioned by Maria, we need to use one of these options, right?

```
1. Still using cp, with that option we are a bit blind with the possible error but still we can work with it:
a) Set the env variable EOS_FUSE_DEBUG=1 and EOS_FUSE_LOGLEVEL=7 in the file /etc/sysconfig/eosd_env

2. Use eoscp or xrootcp:
a) Set the env variable XRD_LOGLEVEL=Dump
b) would be better use the subprocess module in order to get also the error, if there is any.

Then for both options, I would like that you collect the time that is taking to run the writing command, so you calculate the start_time and the end_time, before and after it.
```

thongonary:
> but then how are you going to get in the if block on line 1375 or 1383 when the copy fails?

It was my way of commenting it out. After 5 failed copy attempt they will send an email anyway.

> Also, as mentioned by Maria, we need to use one of these options, right?

Yes. I haven't had time to implement it.

thongonary:
@sharad1126 please add the following:
1. Check file stat of both source and destination file
2. Use `subprocess` instead of `os`, as instructed by Maria to save the log.

PR #446, opened 2019-09-03
add dynamo in components ; soft or not?

vlimant:
#### Status
ready

#### Description
add dynamo in check

y


sharad1126:
I think it is better to keep it as a soft component. @thongonary what do you think?

haozturk:
AFAIK, we do not use dynamo anymore, so closing this PR.

Issue #454, opened 2019-09-30
requests.packages.urllib3 module import error


sharad1126:
Looks like importing JIRA is giving issues. 
```
File "Unified/stagor.py", line 11, in <module>
    from htmlor import htmlor
  File "/data/unified/WmAgentScripts/Unified/htmlor.py", line 11, in <module>
    from JIRAClient import JIRAClient
  File "/data/unified/WmAgentScripts/Unified/JIRAClient.py", line 2, in <module>
    import jira
  File "/usr/lib/python2.7/site-packages/jira/__init__.py", line 10, in <module>
    from jira.client import Comment  # noqa: E402
  File "/usr/lib/python2.7/site-packages/jira/client.py", line 6, in <module>
    from requests.auth import AuthBase
  File "/usr/lib/python2.7/site-packages/requests/__init__.py", line 58, in <module>
    from . import utils
  File "/usr/lib/python2.7/site-packages/requests/utils.py", line 32, in <module>
    from .exceptions import InvalidURL
  File "/usr/lib/python2.7/site-packages/requests/exceptions.py", line 10, in <module>
    from .packages.urllib3.exceptions import HTTPError as BaseHTTPError
  File "/usr/lib/python2.7/site-packages/requests/packages/__init__.py", line 95, in load_module
    raise ImportError("No module named '%s'" % (name,))
ImportError: No module named 'requests.packages.urllib3'
```

https://cms-unified.web.cern.ch/cms-unified/logs/transferor/2019-09-30_17:10:58.log
https://cms-unified.web.cern.ch/cms-unified/logs/htmlor/2019-09-30_17:00:35.log
https://cms-unified.web.cern.ch/cms-unified/logs/stagor/2019-09-30_16:22:34.log

sharad1126:
@thongonary should we disable JIRA client for the time being?

https://github.com/CMSCompOps/WmAgentScripts/blob/master/JIRAClient.py

thongonary:
@sharad1126  please go ahead.

sharad1126:
After analysing it looks like a dependency error from requests, urllib3 or chardet. 
The current versions are:
requests: 2.22.0
urlli3: 1.25.6
chardet:3.0.4

To solve the issue, I have manually downgraded the chardet to 3.0.2 which doesn't gives the error as of now. But I think as the puppet cycle runs on the machines, it might update the versions again which might cause trouble. 

https://its.cern.ch/jira/browse/CMSVOC-100

[...]

PR #455, opened 2019-10-04
Quick fix for JIRA import failures

thongonary:
Fixes #<GH_Issue_Number>
https://github.com/CMSCompOps/WmAgentScripts/issues/454

#### Status
ready

#### Description
Reinstalling the packages whenever `import jira` failed. This might be a dangerous practice, but a quick and dirty solution for now -- unless we have a more elegant way to handle this.

YES

What it does is essentially
```
sudo yum remove python-requests python-urllib3
sudo pip uninstall urllib3 requests
sudo pip install urllib3==1.25.6 requests==2.22.0 chardet==3.0.2
```

@amaltaro @drkovalskyi @sharad1126

### DIFF
--- a/JIRAClient.py
+++ b/JIRAClient.py
@@ -1,6 +1,25 @@
+from subprocess import PIPE,Popen
 import sys,os,json
-import jira
 import time
+
+# JIRA requires a peculiar combination of package versions sometimes
+try:
+    import jira
+except ImportError as e:
+    try:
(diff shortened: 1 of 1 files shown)


sharad1126:
@thongonary I think instead of uninstalling the packages, let's do a force-reinstall if import jira fails. something like:
```
sudo pip install --upgrade --force-reinstall requests
```
as most of the issues are coming due to requests package

PR #462, opened 2019-11-06
change of dbs apis

vlimant:
from @vkuznet suggestions.
working in progress

### DIFF
--- a/utils.py
+++ b/utils.py
@@ -4985,7 +4985,7 @@ def _findParent( dataset ):
 def setFileStatus(file_names, validate=True):
     dbswrite = DbsApi(url=dbs_url_writer)
     dbsapi = DbsApi(url=dbs_url)
-    files = dbsapi.listFileArray(logical_file_name = file_names, detail=True)
+    files = dbsapi.listFiles(logical_file_name = file_names, detail=True)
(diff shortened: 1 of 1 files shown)


utils.py
```diff
@@ -3350,6 +3353,7 @@ def getDatasetFileArray( dataset, validFileOnly=0, detail=False, cache_timeout=3
 
 def getDatasetFileFraction( dataset, files):
     # VK TODO: can be replaced with dataset summary API which returns total number of evnts, files
+    # JRV: this requires to count only valid files.
```
vkuznet: Jean-Roch, as far I remember DBS summary API counts only valid files. It is definitely you should consult with Yuyi which API to use here.

utils.py
```diff
@@ -3781,6 +3786,7 @@ def _getDatasetBlocksFraction(url, dataset, complete='y', group=None, vetoes=Non
         vetoes = ['MSS','Buffer','Export']
 
     # VK TODO: replace with call to get list of blocks for a given dataset
+    # JRV: need only the blocks with valid files.
```
vkuznet: I think DBS provide a way to get list of blocks with valid files, please consult with Yuyi

PR #464, opened 2019-11-14
No announce with duplicates

vlimant:
Fixes https://cms-unified.web.cern.ch/cms-unified/showlog/?search=task_HIG-RunIISummer19UL17wmLHEGEN-00496 output to be set valid ("announce") with a duplicates and filemismatch


sharad1126:
yeah, that's wight. would be helpful. I will test and merge it.

PR #466, opened 2019-11-20
APIs to campaign configuration in ReqMgr2; updated Ops campaign script

amaltaro:
Fixes https://github.com/dmwm/WMCore/issues/9310

#### Status
not-tested

#### Description
Summary of changes is:
* migrate from optparse to argparse (the former is not supported in python3)
* client APIs to interact with ReqMgr2 `campaignconfig` API.
* function to parse MongoDB campaign records and return WMCore formatted campaigns
* minor reorganization of the current campaignsConfiguration.py script
* then added a few extra methods such that we can reproduce MongoDB actions against CouchDB (ReqMgr2)
* updated `batchor` module to create (and delete) the same campaigns it updates in MongoDB.
* updated `closor` to delete campaigns in CouchDB as well

The usage of that script remains the same, same options and so on. The only difference is that it will try to make the same changes on two different backends, first MongoDB and then central CouchDB (production CMSWEB).

It uses the same logic - to convert unified campaign to wmcore campaign - as implemented in this script:
https://github.com/dmwm/WMCore/blob/master/bin/adhoc-scripts/parseUnifiedCampaigns.py

#### Is it backward compatible (if not, which system it affects?)
Sort of, shouldn't change anything in the unified ecosystem.

#### Related PRs
none

#### External dependencies / deployment changes
none

#### Mention people to look at PRs
none yet

PR #468, opened 2019-11-21
Parse correct campaign config in batchor, for wmcore format

amaltaro:
Complement to #466 (read a bug fix for)

#### Status
not-tested

#### Description
Sharad reported this batchor crash over slack
```
File "Unified/batchor.py", line 154, in <module>
    batchor(url)
  File "Unified/batchor.py", line 85, in batchor
    wmcoreCamp = parseMongoCampaigns(campaign)
  File "/data/unified/WmAgentScripts/campaignAPI.py", line 132, in parseMongoCampaigns
    conf[wmKey] = rec.get(uniKey, conf[wmKey])
AttributeError: 'unicode' object has no attribute 'get'
```

The previous PR was passing a plain string (campaign name) to the function to parse a unified campaign dictionary to a wmcore dictionary.
So this PR should be passing the correct data structure (with the `name` key/value pair), which will then be converted to a WMCore campaign schema.

#### Is it backward compatible (if not, which system it affects?)
yes

#### Related PRs
Bugfix for #466 

#### External dependencies / deployment changes
none

#### Mention people to look at PRs
@sharad1126 @thongonary could one of you please review it and if possible push it in in order to fix the batchor bug? Thanks

PR #471, opened 2019-11-25
Accept list with a single campaign config in the createCampaignConfig method

amaltaro:
Fixes #466 and #468

#### Status
not-tested

#### Description
There is still a problem in the code, sorry!
Problem is that when we parse the MongoDB campaign and create a WMCore one, the `parse` method returns a list of dictionaries. We were then passing it over to the `createCampaignConfig` method that was dealing with it as a dictionary (but it was a list of one dict).

This should fix that problem. If a list of multiple campaigns (multiple dicts) is passed in, the createCampaignConffig will report an error.

yes

#### Related PRs
Said in the first line, it's a bug fix for #466 and #468

none

@sharad1126 hoping it to be the last PR now.

### DIFF
--- a/campaignAPI.py
+++ b/campaignAPI.py
@@ -27,6 +27,11 @@ def createCampaignConfig(docContent, url=reqmgr_url):
     :param docContent: dictionary with the campaign content
     :return: a boolean whether it succeeded (True) or not (False)
     """
+    if isinstance(docContent, list) and len(docContent) > 1:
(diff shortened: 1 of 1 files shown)


sharad1126:
@amaltaro thanks for looking into this.
I have already merged your PR for the .idea thing, there is a commit for that here as well. (https://github.com/CMSCompOps/WmAgentScripts/pull/470)

amaltaro:
Sorry, I created the PR in sort of a rush and forgot to review it before pushing the button. Let me rebase it.

PR #477, opened 2019-12-13
Use single dbsApi/dbsApiWriter instance

vkuznet:
Fixes #418 

#### Status
In development

#### Description
Replace multiple dbs Apis instances with single one initialized for this module

YES

#### Related PRs
replaces #418


haozturk:
I see that the related GH issue is closed and as this is a quite old PR, I am closing it. Feel free to re-open it if you find it necessary.

PR #479, opened 2020-01-23
TaskChain to StepChain conversion improvement

drkovalskyi:
This PR allows to do TaskChain to StepChain conversion based on the expected CPU inefficiency, which is computed as the average of wall-clock time CPU inefficiency. The efficiency estimation is most likely too conservative since some steps can benefit from extra cores even if max number of cores is not optimal. For now the threshold is set to 75%.

### DIFF
--- a/utils.py
+++ b/utils.py
@@ -6628,6 +6628,7 @@ class workflowInfo:
 
         self.summary = None
 
+        self.UC = unifiedConfiguration()
 
 
 
@@ -6677,16 +6678,13 @@ class workflowInfo:
         timeInfo = dict()
         if 'Chain' in self.request['RequestType']:
             base = self.request['RequestType'].replace('Chain','')
-            itask=1
-            carry_on = {}
-            while True:
+            for itask in range(1, self.request['TaskChain'] + 1):
                 t = '%s%d'%(base, itask)
-                itask+=1
                 if t in self.request:
                     task = self.request[t]
                     timeInfo[t] = {
                         'tpe':task.get('TimePerEvent')/task.get('FilterEfficiency',1),
-                        'cores':task.get('Multicore',1)
+                        'cores':task.get('Multicore', self.request['Multicore'])
                     }
                 else:
                     break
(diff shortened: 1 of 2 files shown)


utils.py
```diff
@@ -6673,8 +6673,42 @@ def isGoodForNERSC(self, no_step=False):
...
+            itask=1
+            carry_on = {}
+            while True:
+                t = '%s%d'%(base, itask)
+                itask+=1
+                if t in self.request:
+                    task = self.request[t]
+                    timeInfo[t] = {
+                        'tpe':task.get('TimePerEvent')/task.get('FilterEfficiency',1),
+                        'cores':task.get('Multicore',1)
```
amaltaro: Dima, `Multicore` is an optional argument and it might be present only at the top level (even though it could be that McM always set them in each task). So you might want to change this line from
```
'cores':task.get('Multicore',1)
```
to
```
'cores':task.get('Multicore', self.request['Multicore'])
```

utils.py
```diff
@@ -6673,8 +6673,42 @@ def isGoodForNERSC(self, no_step=False):
         #good = good &
         return good
 
+    def getTimeInfoForChain(self):
+        timeInfo = dict()
+        if 'Chain' in self.request['RequestType']:
+            base = self.request['RequestType'].replace('Chain','')
+            itask=1
+            carry_on = {}
+            while True:
```
amaltaro: I'm always scared with these "While True", so you could replace it by something like:
```
for itask in range(1, self.request['TaskChain'] + 1):
```
this will yield the exact numbers of Tasks in the workflow.

utils.py
```diff
@@ -6673,8 +6673,42 @@ def isGoodForNERSC(self, no_step=False):
...
+        efficiency = 0
+        max_ncores = 1
+        for i,info in time_info.items():
+            totalTimePerEvent += info['tpe']
+            efficiency += info['tpe']*info['cores']
+            if info['cores']>max_ncores: max_ncores = info['cores']
+        # print "Total time per event for TaskChain: %0.1f" % totalTimePerEvent
+        efficiency /= totalTimePerEvent*max_ncores
+        # print "CPU efficiency of StepChain with %u cores: %0.1f%%" % (max_ncores,efficiency*100)
+        acceptable_efficiency = efficiency>0.75
```
thongonary: We could also set this 0.75 to be configurable in https://github.com/CMSCompOps/WmAgentScripts/blob/master/unifiedConfiguration.json
E.g. adding an entry in unified config
```
 "efficiency_threshold_for_stepchain" : {
   "value" : 0.75,
   "description" : "Minimum efficiency threshold for step chain conversion"
 },
```
and in this `utils.py` we do
`acceptable_efficiency = efficiency>UC.get("efficiency_threshold_for_stepchain")`
So that one year later whoever maintaining Unified can just change the config file without tracing back to the codebase.

utils.py
drkovalskyi: done

utils.py
drkovalskyi: done

utils.py
drkovalskyi: done

PR #480, opened 2020-01-25
protection against missing information

drkovalskyi:
Apparently in some cases the information may not be available to compute what we need or it can be of a  wrong type. So in case of an exception set a flag that "efficiency" is not good to convert.


vlimant:
@drkovalskyi several lines are over-indented IMO

PR #482, opened 2020-01-29
only pick requests for dataops and anaops

vlimant:
with regards to 
https://hypernews.cern.ch/HyperNews/CMS/get/comp-ops/4856.html 
https://its.cern.ch/jira/browse/CMSCOMPPR-11171 

https://cms-unified.web.cern.ch/cms-unified/showlog/?search=ReReco-Run2018A-SingleMuon-12Nov2019_UL2018-0001&module=transferor&limit=500&size=10000

> The input is either fully in place or getting in full somewhere with [2000338]
> latches on existing transfers, and nothing else, settin staging
> setting pgunnell_Run2018A-v1-SingleMuon-12Nov2019_UL2018_1064p1_191114_160444_8742 status to staging

only request for AnalysisOps and DataOps should be considered as good source for production

### DIFF
--- a/utils.py
+++ b/utils.py
@@ -3861,11 +3861,10 @@ def getBetterDatasetDestinations( url, dataset, only_blocks=None, group=None, ve
 
     print len(all_block_names),"blocks"
 
+    conn = make_x509_conn(url)
     items = None
-    while items == None:
+    while items is None:
         try:
-            conn = make_x509_conn(url)
-            #conn  =  httplib.HTTPSConnection(url, cert_file = os.getenv('X509_USER_PROXY'), key_file = os.getenv('X509_USER_PROXY'))
             url = '/phedex/datasvc/json/prod/requestlist?dataset=%s&group=AnalysisOps&group=DataOps'%dataset
             if group:
                 url+='group=%s'%group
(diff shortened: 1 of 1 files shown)


sharad1126:
Thanks, will test and merge this.

sharad1126:
Using the updated URL, phedex doesn't returns the `type: "delete"` requests and shows only transfer requests `type: "xfer"` . So I don't think the same logic of the deletion that is being checked https://github.com/CMSCompOps/WmAgentScripts/blob/b17c8c6f6c23d1232d8244bca5c4ff0baec21c64/utils.py#L3882-L3893 will work.

sharad1126:
May be this - https://cmsweb.cern.ch/phedex/datasvc/json/prod/subscriptions?dataset=/SingleMuon/Run2018A-v1/RAW&create_since=1 can help us. It shows the existing subscriptions and the groups so we don't need any maths if there are deletions that happened or according to timestamps.

vlimant:
subscriptions does not work for this

sharad1126:
Thanks for adding the type:delete. I also think that in https://github.com/CMSCompOps/WmAgentScripts/blob/06118e25c6dab70444e6fb5902038cb2668147b6/utils.py#L3980 there is one extra '%' after %s which takes the code into the exception

PR #483, opened 2020-01-29
hook to close all stopped campaigns

vlimant:
so as to remove the man in the middle, unified can look in mcm and disable all stopped campaigns

before merging, one needs to uncomment (and test) the update_one command

### DIFF
--- a/utils.py
+++ b/utils.py
@@ -1456,6 +1456,19 @@ class campaignInfo:
                             reg = black[0:-1]
                             self.campaigns[c]['parameters']['SiteBlacklist'].extend( [site for site in (SI.all_sites) if site.startswith(reg)] )
                             #print self.campaigns[c]['parameters']['SiteBlacklist']
+    def close_stopped(self):
+        # get all GO campaigns
+        all_go = [c['name'] for c in self.db.find({"go":True})]
+        # get all stopped campaigns
+        from McMClient import McMClient
+        mcm = McMClient(dev=False)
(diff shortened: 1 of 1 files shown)


sharad1126:
We already have a policy to close the campaigns. We have the script remove_campaigns to remove it from the dB, and copy it into archive json file.

vlimant:
that still requires a human in the loop. this PR removes the ops-human

sharad1126:
I went through the code and this will cause the same issue i.e. not removing the campaign from the json file. Therefore, the last time we discussed about this, we made the remove_campaigns.py which actually removes the campaign from the mongodb and copies it to the archive json file so operator doesn't has sees those campaigns in the campaigns.json file.

vlimant:
it could be as easy as dumping only go=true campaigns in campaigns.json and the go=false in archive_campaigns.json, or add an "archive" field somewhere (although as I said in the first place, there is no need for the archive file, and despite my comments and arguments it was put in place, barring the way to deployement of this automatic procedure. FYI @drkovalskyi ) there is no justification to increasing human intervention, when actually people claim they have too much on their hands

sharad1126:
During operations, I felt it to be too much to go through campaigns in campaigns.json even if the campaigns aren't used so I worked with @optimisticcynic who agreed with this for managing campaigns. I leave it on the L3s (@jenimal @z4027163) and @scarletnorberg as she will be the one who will be handling campaigns after @optimisticcynic leaves.

PR #484, opened 2020-01-31
force-complete and bypass directly from jira

vlimant:
with regards to https://its.cern.ch/jira/browse/CMSCOMPPR-11343 if one wants to avoid having to operate manually, see this PR.
adding a "cmsunified please do force-complete wfname" or "cmsunified please do bypass wfname" will result in the desired action

### DIFF
--- a/Unified/checkor.py
+++ b/Unified/checkor.py
@@ -158,7 +158,7 @@ def checkor(url, spec=None, options=None):
                     keyword = c.body[(c.body.find(force_complete_jira_string)+len(force_complete_jira_string)):].split()[0]
                     if keyword and user in actors:
                         print user,"is force-completing", keyword,"from JIRA"
-                        bypasses.update( keyword )
+                        bypasses.extend( keyword )
                         overrides[user].extend( keyword )
                     break
         bypass_jira_string = "cmsunified please do bypass"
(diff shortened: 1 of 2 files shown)


sharad1126:
@vlimant I tried to test the PR and got the following error:
```
sagarwal is bypassing pgunnell_Run2018D-v1-EGamma-12Nov2019_UL2018_1064p1_200113_211959_1205 from JIRA
  File "Unified/checkor.py", line 1573, in <module>
    checkor(url, spec, options=options)
  File "Unified/checkor.py", line 176, in checkor
    bypasses.update( keyword )
AttributeError: 'list' object has no attribute 'update'
```
I think it should be bypasses.extend(keyword) instead of https://github.com/CMSCompOps/WmAgentScripts/blob/77df324fac287324e9777a56a0fb992a9691f26b/Unified/checkor.py#L175

Unified/checkor.py
```diff
@@ -158,7 +158,7 @@ def time_point(label="",sub_lap=False, percent=None):
                     keyword = c.body[(c.body.find(force_complete_jira_string)+len(force_complete_jira_string)):].split()[0]
                     if keyword and user in actors:
                         print user,"is force-completing", keyword,"from JIRA"
-                        bypasses.update( keyword )
+                        bypasses.extend( keyword )
```
vlimant: because 
>>> l=range(10)
>>> l.extend('keyword')
>>> l
[0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 'k', 'e', 'y', 'w', 'o', 'r', 'd']
>>>

it should be append, and not extend

>>> l=range(10)
>>> l.append('keyword')
>>> l
[0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 'keyword']

you are lucky we have been through https://github.com/CMSCompOps/WmAgentScripts/commit/f94c63ce4ade9c505386e8d1ca622654ec14bcc6 in the past, otherwise all current workflows in assistance would have been bypassed

sharad1126: thank you for checking this. would have been really bad.

PR #489, opened 2020-02-06
Remove legacy dashssb

sharad1126:
Fixes #487 #488 

#### Status
tested  

#### Description
<Description of the changes proposed.>

#### Is it backward compatible (if not, which system it affects?)
NO

#### Related PRs
#458 

#### External dependencies / deployment changes
dash-ssb*, dash-cms*

#### Mention people to look at PRs
@z4027163 @vlimant @amaltaro

PR #490, opened 2020-02-07
htmlor code with legacy mistakes

sharad1126:
Fixes errors in htmlor logs:
```
sort: cannot read: 6: No such file or directory
grep: write error: Broken pipe
```

#### Status
tested

#### Description
htmlor module logs shows errors due to some mistakes in code. 

NO

#### Related PRs
#489 

sort -f

@z4027163

### DIFF
--- a/Unified/htmlor.py
+++ b/Unified/htmlor.py
@@ -844,7 +844,7 @@ Worflow through (%d) <a href=logs/closor/last.log target=_blank>log</a> <a href=
 %s
 </pre>
 """%(os.getenv('USER'),
-     os.popen('acrontab -l | grep -i unified | grep -v \#  |sort -f 6').read()))
+     os.popen('acrontab -l | grep -i unified | grep -v \#  |sort -f ').read()))
 
 
     per_module = defaultdict(list)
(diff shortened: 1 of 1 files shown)


vlimant:
interesting. you actually want "-k 6" instead of "-f 6"

sharad1126:
thanks @vlimant will make the changes

sharad1126:
@vlimant can you also please check from where is the following line coming in the logs and what needs to be done?
```
The dotted field 'vocms0815.cern.ch' in 'data.Negotiation.vocms0815.cern.ch' is not valid for storage.
```

vlimant:
this line is coming from https://github.com/CMSCompOps/WmAgentScripts/commit/ae818b12409fdcf54174bee9c3707d948355a93c#diff-04d2c9518498da64bfa9db44c6211645

where documents are cached in mongodb. When the key has a "." mongodb rejects it and it reverts to using file based caching (small fraction of everything cached).

the way around it is to find an efficient/transparent way of doing replace(".","___dot___") (THAT actually might just be the solution )

vlimant:
actually, now remembering better, this is coming from "." nested in the fields of the json being pushed to mongo https://github.com/CMSCompOps/WmAgentScripts/commit/bdc25648cac6c61c1605edc09180006e3381ada2 will not solve it. the error is harmless, and you probably have better to do

sharad1126:
thanks for the quick responses. I will merge this PR for now. Thanks again.

PR #491, opened 2020-02-11
checkTransferLag fetch collapse=n when dataset sub is available

### LINKED ISSUE #450: phedex url issue for checkTransferLag
@nsmith- @dmielaikaite @nataliaratnikova can you please have a look?

```http://cmsweb.cern.ch/phedex/datasvc/json/prod/subscriptions?request=1687295&create_since=1553024209.35001&collapse=n```

The url after appending '&collapse=n' isn't working or taking too much time to load. 

@thongonary  FYI

### PR DESCRIPTION
**Pull request by z4027163**
Fixes #450

#### Status
Tested for stuckor module

#### Description
checkTransferLag looks into 'collapse=n' for block substription only when dataset subscription is available in 'collapse=y'.

Yes

#### Related PRs
N/A

NO

@sharad1126 @drkovalskyi

### DIFF
--- a/utils.py
+++ b/utils.py
@@ -3514,14 +3514,13 @@ def try_checkTransferLag( url, xfer_id , datasets=None):
 
     for item in  result['phedex']['dataset']:
         if 'subscription' in item:
-             add_collapse=True
-             break
+            add_collapse=True
+            break
     if add_collapse:
         subs_url+='&collapse=n'
         r1=conn.request("GET",subs_url)
         r2=conn.getresponse()
         result = json.loads(r2.read())
-
 
     now = time.mktime( time.gmtime() )
     #print result
--- a/utils.py
+++ b/utils.py
@@ -3512,7 +3512,7 @@ def try_checkTransferLag( url, xfer_id , datasets=None):
     #only check '&collapse=n' when there is dataset level subscription
     add_collapse=False
 
-    for item in  result['phedex']['dataset']:
+    for item in result['phedex']['dataset']:
         if 'subscription' in item:
             add_collapse=True
             break
@@ -3539,7 +3539,7 @@ def try_checkTransferLag( url, xfer_id , datasets=None):
                 result = json.loads(r2.read())
                 break
 
-    for item in  result['phedex']['dataset']:
+    for item in result['phedex']['dataset']:
         if 'subscription' not in item:
             #print "sub"
             loop_on = list(itertools.chain.from_iterable([[(subitem['name'],i) for i in subitem['subscription']] for subitem in item['block']]))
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by vlimant** [comment]
just for my education, what's the actual issue you are trying to solve ?

**Comment by sharad1126** [comment]
If I am not wrong, the collapse=n doesn't always work for this request and ends up giving no json error in critical logs. The last comment by Nick I see on the chat was the following:
```
so I eventually found the issue these requests are failing is actually due to a timeout in the phedex backend trying to do the SQL query. It turns out that queries to subscriptions with only the request ID as a key end up being very expensive due to the way the query was written.  We might have to find another way to query the status, I’ll put together a recipe using a sequence of other datasvc commands
```
 So I am not sure, what happened after that conversation as that was 1 month ago.

**Comment by vlimant** [comment]
are you talking about 

> running the check on 2077436
> cache hit phedex_nodes
> trying with an earlier date than 1579102650.5 for 2077436
> {u'phedex': {u'request_date': u'2020-02-12 10:00:46 UTC', u'request_timestamp': 1581501646.97392, u'call_time': 0.00492, u'dataset': [], u'instance': u'prod', u'request_call': u'subscriptions', u'request_url': u'http://cmsweb.cern.ch:7001/phedex/datasvc/json/prod/subscriptions', u'request_version': u'2.5.0pre3-comp'}}
> withouth an update, we are in some trouble.
> Failed transfer status check on 2077436

for example ?

**Comment by vlimant** [comment]
what is the related github issue ?

**Comment by z4027163** [comment]
@vlimant 
The error comes from stuckor:
>  fatal execption in checkTransferLag for 2085796
> [u'/MET/Run2017E-v1/RAW', u'/MET/Run2017C-v1/RAW']
> No JSON object could be decoded
which is caused https://github.com/CMSCompOps/WmAgentScripts/blob/master/utils.py#L3508 when r1 does not exist.

**Comment by sharad1126** [comment]
@nsmith- can you please also have a look here. If you remember we discussed about this on the slack channel 1-2 months ago about the error happening while getting this json.

**Comment by nsmith-** [comment]
So if I recall, the error was that sometimes the query timed out.  On the datasvc side, the SQL query is not so well designed for the case `collapse=n`.  Having unified not use this flag unless necessary might help a bit in this regard, but certainly its still possible in the case where there is a dataset-level subscription and another query needs to be made.

**Comment by sharad1126** [comment]
@z4027163 where are we with this PR? can you please reply to Nick's comment.

**Comment by sharad1126** [comment]
@z4027163  we need to resolve this out soon. critical logs are full with this exception, eg:
```
fatal execption in checkTransferLag for 2135271
[u'/Cosmics/Run2018D-v1/RAW']
No JSON object could be decoded
```

**Comment by z4027163** [comment]
@nsmith- Sorry for the late reply. This pull request does another query with flag 'collapse=n'  if the dataset level subscription in default log is found.

**Comment by sharad1126** [comment]
@z4027163 as per @nsmith- 's last comment on slack:
```
uh well if you look at the docs, collapse is not an option for the requestlist API call https://cmsweb.cern.ch/phedex/datasvc/doc/requestlist
```
I think we need to remove https://github.com/CMSCompOps/WmAgentScripts/blob/master/utils.py#L3389

haozturk:
The functions and modules in question were deprecated, so closing this PR.

PR #495, opened 2020-02-27
Module mappor

vlimant:
separating the unified overflow mapping from the equalizor module.

@drkovalskyi


sharad1126:
@vlimant can I please ask you to explain where is the mappor module called because I don't see it being called in equalizor and no other file as per this PR has been modified.

sharad1126:
ohhh ok my bad. I see you are using the cache here.

sharad1126:
@vlimant can you please also tell the acron job you want to run with this module as I assume you would like to run it as an acron so mapping works and cache is updated everytime or am I missing anything?

vlimant:
you can set it in cron the way to see fit

sharad1126:
@vlimant a little confused by the hlt option. Shouldn't we be allowing hlt overflow when we have site in production and either the input data is at CERN disk or no input is needed(considering it matches the cpu requirements). I mean why do we have the option for hlt explicitly?

z4027163:
I am wondering what's the benefits of separating mapping from equalizor?  Does it mean we can run mappor less frequent than equalizor, therefore reducing the loads?

vlimant:
this is a step towards having the mapping provided from "outside", and controlling the mapping independently of running equalizor

PR #496, opened 2020-02-29
ACDC priority boost in terms of WF size

z4027163:
ACDC priority increase for 20% if the ACDC has more than 500 jobs.
@vlimant @jenimal

### DIFF
--- a/Unified/actor.py
+++ b/Unified/actor.py
@@ -42,7 +42,9 @@ def singleRecovery(url, task, wfi, actions, do=False):
     #Max priority is 1M
     #low boost for #jobs > 500
     heavyacdc = False
-    failjobs = wfi.getFailedJobs(task)
+
+    task_name = tasl.split("/")[-1]
+    failjobs = wfi.getFailedJobs(task_name)
     if failjobs and failjobs>500:
         heavyacdc = True
 
--- a/Unified/actor.py
+++ b/Unified/actor.py
@@ -43,8 +43,7 @@ def singleRecovery(url, task, wfi, actions, do=False):
     #low boost for #jobs > 500
     heavyacdc = False
 
-    task_name = tasl.split("/")[-1]
-    failjobs = wfi.getFailedJobs(task_name)
+    failjobs = wfi.getFailedJobs(task)
     if failjobs and failjobs>500:
         heavyacdc = True
 
--- a/Unified/actor.py
+++ b/Unified/actor.py
@@ -29,7 +29,7 @@ def singleRecovery(url, task, initial, actions, do=False):
         "OriginalRequestName" : initial['RequestName'],
         "OpenRunningTimeout" : 0
     }
-    copy_over = ['PrepID','Campaign','RequestPriority', 'TimePerEvent', 'SizePerEvent', 'Group', 'Memory', 'RequestString' ,'CMSSWVersion','TotalEstimatedJobs']
+    copy_over = ['PrepID','Campaign','RequestPriority', 'TimePerEvent', 'SizePerEvent', 'Group', 'Memory', 'RequestString' ,'CMSSWVersion']
     for c in copy_over:
         if c in initial:
             payload[c] = copy.deepcopy(initial[c])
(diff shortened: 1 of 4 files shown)


vlimant:
it looks like the condition is based on the number of job on the initial workflow, and not on the acdc itself.

z4027163:
@vlimant I thought the initial wf is made from vocms0113 via the console. So it should be in the format of ACDC, but with most of the wf information copied from the original WF only with the failed blocks as input.

vlimant:
TotalEstimatedJobs is from the original workflow, and not how many jobs will land in the acdc

z4027163:
The information of this "original workflow" is from vocms0113, which I think is already the prototype of the ACDC workflow (which is produced by console when we submit acdc request). So I think the TotalEstimatedJobs actually means the jobs in this prototype acdc workflow.

vlimant:
it is not

vlimant:
one could retrieve how many events will be in each acdc, at least at the base task level.

z4027163:
When I check the wfi.request info for ACDC workflow, the TotalEstimatedJobs shows the number of jobs of ACDC. So at some step, it calculates the job number of the ACDC in unified? 
I checked the log file, the initial input workflow name is already in the format of "cmsunified_ACDC_*".

z4027163:
@vlimant I am now using outputDataset lumi to get the failed jobs for each of the task, which is the estimated as the acdc input jobs. Let me know if there is faster/easier way for unified to get that information.

z4027163:
@vlimant I did some tests to check the number of failed job calculations in comparison to the actual ACDC jobs, the result seems reasonable (although some low values may differ a bit due to rounding). The splitting consideration part can not be tested, as this happens after the worflow submission.  If there is no major issue, I would like to merge it today.

vlimant:
I have concerns (@drkovalskyi ) about the fact that downgrading the priority is going to create backlog for operation by not having the ACDC running right away and things clearing off (both for operation, space, ...) 

It looks like there is code duplication in what is in this PR ; have you checked this carefully?

Instead of implementing something with a build in approximation on how heavy the ACDC will be, it would be more accurate to check on the number of jobs, post-assignment https://github.com/CMSCompOps/WmAgentScripts/blob/master/Unified/actor.py#L654, or better even to implement jobrouting/wmagent rules to limit the amount of ACDC jobs with high priority from a global point of view.

sharad1126:
@vlimant as per the L1s and L2s concerns of ACDCs overcoming the requests in block1, we have to do the downgrade of priority for ACDCs. Would be great if you have a suggestion for this on how to implement it.

vlimant:
I just made suggestions.

This whole episode started with the 2019/2020 Xmas break when hundreds of ACDC accumulated (some over trivial failures/recoveries) : so better Ops work on the automated ACDC (as I already mentioned a couple of times) so that most ACDC get out without human intervention.

Is this still an issue today ? no it is not (see gwmsmon)
Do ACDC have to get done no matter what ? yes (except in case of code 71305 https://twiki.cern.ch/twiki/bin/view/CMSPublic/JobExitCodes)
Are we wasting time and energy on this ? IMO yes. but good that @z4027163 is getting a hold of the internals

sharad1126:
Totally agreed. Thanks.

drkovalskyi:
Thanks @vlimant . It's a valid concern. Taking it into account and the fact that we don't have solid evidence that we have an issue with too many ACDCs jobs causing a priority inversion, it was decided at the last CompOps coordination to postpone any changes to the production configuration. We still should have an option to make the change if necessary.

jenimal:
Agreed on the automation of the ACDC's,  that really should be our priority.  That said, @Todor and I spent time looking at this and the limits we have put in will send the majority of the ACDC's through quickly and just prevent those occasional HUGE ACDC's from taking over production at a high priority when there is a system issue.  If these changes of sending the large ACDC's through with only a slightly higher priority than the parent workflow is ready to go we should just implement it and continue on with the work of automated ACDC, which is something that was supposed to be worked on with the console so we should give our support to T&I to get that done.

z4027163:
@vlimant I tried to get information from post-assigned acdc, but the job number information is not provided. From Thong, the number of jobs is calculated in WMcore. So I don't think that I could get the job number directly from submitted workflows in actor, and it may not be a good idea to change the priority outside this module.  (if I am wrong, I would like to have a talk with you about the details).

Also I think estimating the acdc number is fine, it won't be accurate when job number is low, but our threshold is set to be 500, the estimation shouldn't be far off at that point, and it could serve the purpose of preventing large ACDCs from getting high priority.

Issue #501, opened 2020-03-11
monit is unstable : failed to get site configuration


vlimant:
#492 lead to failures in getting site configuration https://github.com/CMSCompOps/WmAgentScripts/commit/3334d6925ba24af3315dd152978654091cb9767f

sharad1126:
@vlimant i have already a email thread open for this with the site support team and MonIT team.

sharad1126:
Can you please tell why you commented out the send email? It’s good to know when it failed so we can get this fixed from MONIT team.

sharad1126:
Just realised it's due to https://cern.service-now.com/service-portal/view-outage.do?n=OTG0054760.

vlimant:
bombing one's inbox with thousands of emails is not really the way to go, is it. removing that was mandatory. having unified broken due to this is not even a good thing (the last source of site status information was far more stable overall), hence #502 

https://cms-unified.web.cern.ch/cms-unified/showlog/?search=critical&module=doccache was/is the source of information to keep an eye on

vlimant:
should the site status/monit be put in a critical component @drkovalskyi @z4027163 ?
it would then have to be put under componentInfo

sharad1126:
I agree, but I didn't know that it was bombarding and that's why I removed the comment.

I think even if the doccache exists, still it showed no document available for other metrics so I am not sure even putting this one in doccache is going to help a lot.

sharad1126:
@vlimant I thought that the cache was supposed to use the last available doc for the ssb metrics if the ssb fetch is failed according to the last PR #502  but unfortunately looks like that isn't the case. assignor module fails since morning because the site-prod status isn't available.

sharad1126:
From the assignor logs, `2020-04-27_05:40:26.log` ran without issues as the json was present in kibana but the next log `2020-04-27_06:05:25.log` gave the error for the metric and never used the last doc. this is something we need to check why https://github.com/CMSCompOps/WmAgentScripts/blob/master/utils.py#L1884 didn't work as expected.

sharad1126:
The error is as below:

```
cache hit ssb_prod_status
Failed to run function _get_dashbssb with arguments ['sts15min', 'prod_status'] {} for 1/10 times and 5[s] wait (%s). Exception
 Mon Apr 27 06:05:46 2020
 list index out of range
```

vlimant:
the last doc for "ssb_prod_status" is {}
[...]

PR #502, opened 2020-03-11
docCache gets last cached doc

vlimant:
for #501 and more.
for any doc for which the getter is failing, the last doc in cache will be used, even if expired.
a critical message is send anyways on failing to retrieve a fresh doc.


sharad1126:
this sounds good to avoid the error. thanks for the quick response on this.

PR #504, opened 2020-03-19
Feature unified config param

todor-ivanov:
Fixes #<GH_Issue_Number>

#### Status
<In development | not-tested | on hold | ready>

#### Description

<YES | NO>

#### Related PRs
<If it's a follow up work; or porting a fix from a different branch, please mention them here.>

<Does it require deployment changes? Does it rely on third-party libraries?> 

<Always mention people to check and review proposed changed.>

### DIFF
--- a/serviceConfiguration.json
+++ b/serviceConfiguration.json
@@ -1,0 +1,62 @@
+{
+    "mongo_db_url":  {
+        "value": "vocms0274.cern.ch",
+        "description": "The mongodb instance."
+    },
+    "dbs_url":  {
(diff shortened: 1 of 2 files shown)


sharad1126:
@todor-ivanov I thought we decided to make another file and not use the unifiedconfiguration.json

todor-ivanov:
Hi @sharad1126, we are supposed to use the serviceConfiguration.json file. I just added a new commit which is introducing the file, and also, now all the global config variables in utils.py are read from there. So far, the values in the file are consistent with the configuration for the production environment. We may fill in the values for the testing environment on the go.

PR #516, opened 2020-03-30
Add debugging to _get_dashbssb

nsmith-:
Fixes #?

#### Status
In development

#### Description
Wrap json.loads() in a try-except block to print out what the return value was on json decode error

YES

#### Related PRs
Following discussion in email thread

No dependency change

@sharad1126

### DIFF
--- a/utils.py
+++ b/utils.py
@@ -1548,19 +1548,48 @@ def getWMStats(url):
 def get_dashbssb(path_name, ssb_metric):
     return runWithRetries(_get_dashbssb, [path_name, ssb_metric], {})
 
+
 def _get_dashbssb(path_name, ssb_metric):
-    with open('Unified/monit_secret.json') as monit:
+    with open("Unified/monit_secret.json") as monit:
         conf = json.load(monit)
     query = """'{"search_type":"query_then_fetch","ignore_unavailable":true,"index":["monit_prod_cmssst_*","monit_prod_cmssst_*"]}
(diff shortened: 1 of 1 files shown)


sharad1126:
thanks @nsmith- . Also, thanks for formatting the code!

vkuznet:
@nsmith- , @sharad1126 , there is no need to invoke curl as subprocess, instead I suggest to revisit code and replace curl with CMSMonitoring [monit.py](https://github.com/dmwm/CMSMonitoring/blob/master/src/python/CMSMonitoring/monit.py) module. It can query ES/InfluxDB and we'd like to keep single place to access MONIT rather have different implementation. Plus it is python implementation instead of external curl one.

nsmith-:
For sure curl in a subprocess is not a good practice. But let's take this one step at a time and diagnose the problem at hand.

vkuznet:
Then I suggest explicitly mark NOT TO Merge this PR in production. If you need to debug stuff, you may fix the running code, but then push proper implementation into production codebase.

nsmith-:
I just made this as the most convenient way to share a code patch with Sharad. I have no opinion on whether this PR should get merged or not and leave it to unified people to decide.

sharad1126:
Hi @nsmith- , @vkuznet , technically we did plan to do something similar but we cannot use JSONDecodeError as it is python3 and we are still in python2. 
I will be going through the suggestion of using CMSMonitoring monit.py module is it is all in python2 as well. 
Else, we will find some other alternative solution. Thanks.

nsmith-:
Ah sorry, I've been in python3 land for too long!

vkuznet:
Sharad, CMSMonitoring monit.py script is compatible with both versions of python.

PR #518, opened 2020-04-01
Remove more legacy code from closor

sharad1126:
Removed old code from closor

#### Status
not tested

#### Description
```DDMcopies``` is not used anymore in campaigns and therefore shouldn't be used in the code as well. 

no

#### Related PRs
<If it's a follow up work; or porting a fix from a different branch, please mention them here.>

campaigns.json and unifiedconfigurations.json

@z4027163 @vlimant @amaltaro fyi

### DIFF
--- a/Unified/closor.py
+++ b/Unified/closor.py
@@ -552,23 +552,23 @@ class CloseBuster(threading.Thread):
 
                         n_copies = 1
                         destinations=[]
-                        if to_DDM and campaign and campaign in CI.campaigns and 'DDMcopies' in CI.campaigns[campaign]:
-                            ddm_instructions = CI.campaigns[campaign]['DDMcopies']
-                            if type(ddm_instructions) == int:
-                                n_copies = CI.campaigns[campaign]['DDMcopies']
-                            elif type(ddm_instructions) == dict:
-                                ## a more fancy configuration
-                                for ddmtier,indication in ddm_instructions.items():
-                                    if ddmtier==tier or ddmtier in ['*','all']:
-                                        ## this is for us
-                                        if 'N' in indication:
-                                            n_copies = indication['N']
-                                        if 'host' in indication:
-                                            destinations = indication['host']
+                        #if to_DDM and campaign and campaign in CI.campaigns and 'DDMcopies' in CI.campaigns[campaign]:
+                        #    ddm_instructions = CI.campaigns[campaign]['DDMcopies']
+                        #    if type(ddm_instructions) == int:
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by sharad1126** [comment]
@amaltaro if I am not wrong I think we should also remove the variable ```group_spec ``` https://github.com/CMSCompOps/WmAgentScripts/blob/master/Unified/closor.py#L572
and it's use in https://github.com/CMSCompOps/WmAgentScripts/blob/master/Unified/closor.py#L577.

Also, till now the destination for this log was [ ] as it never went into the #L555 so should we also remove the other things from here?

**Comment by amaltaro** [comment]
Yes, I see no reason to keep `group_spec` around, it's always reported as `None`.

**Comment by vlimant** [comment]
what will happen when PPD will ask for the output to go to a specific location?

**Comment by sharad1126** [comment]
@vlimant we have not been using `DDMCopies` parameter since a very long time. The one we use is `toDDM` which is related to unifiedconfiguration.json if I am not wrong and we aren't removing that.

**Comment by sharad1126** [comment]
@amaltaro @vlimant @z4027163 if no one has any more issues, I can merge this.

File: Unified/closor.py [hunk 1]
```diff
@@ -551,30 +551,30 @@ def close(self):
...
+                        #destination_spec = ""
+                        #if destinations:
+                        #    destination_spec = "--destination="+",".join( destinations )
+                        #group_spec = "" ## not used yet 
                         ### should make this a campaign configuration
                         ## inject to DDM when necessary
                         if to_DDM:
                             print "Sending",out," to DDM"
-                            status = pass_to_dynamo( [out], N = n_copies, sites=destinations if destinations else None, group = group_spec if group_spec else None)
+                            status = pass_to_dynamo( [out], N = n_copies, sites=destinations if destinations else None, None)
```
amaltaro, line 577: `destinations` is not initialized inside this if-else block. Either you initialize it as done before `destinations=[]`, or you update this `pass_to_dynamo` line to `sites=None`

sharad1126: I think there is more here that we are missing.` [L531-L551](https://github.com/CMSCompOps/WmAgentScripts/blob/master/Unified/closor.py#L531-L551)` is used now but I am not able to see the transfers that should happen for these if `to_DDM` is `true`.

haozturk:
As closor module has changed lately, this PR has some conflicts. I did not check the details, but we will revisit this legacy code in the near future. So, I prefer closing this PR and opening a new one.

PR #528, opened 2020-04-14
Fix KeyError exception while fetching failures from wmstatsserver

amaltaro:
Fixes  (is there any issue for this?)

#### Status
not-tested

#### Description
Handle cases where one of those keys (e.g. `status`) is missing.

Just a note, if we have many tasks to be recovered within the same workflow, it appears to me that we would make that wmstatsserver call many times. That query is heavy and it should be avoided as much as we can (thus, single query for the workflow and handle multiple tasks on the client side).

#### Is it backward compatible (if not, which system it affects?)
yes

#### Related PRs
none

#### External dependencies / deployment changes
none

#### Mention people to look at PRs
@sharad1126

PR #529, opened 2020-04-15
lock actor on failure

vlimant:
mentioned in #528.

there's something to be understood on what mechanism is locking based on /eos/cms/store/unified/actor*.lock ; I think it's a legacy not taken care of from 03caf9c12308a3f4a8a13cbdb60163eda2499bed

something needs to be changed then to use moduleLock

### DIFF
--- a/Unified/actor.py
+++ b/Unified/actor.py
@@ -15,6 +15,7 @@ import httplib
 import ssl
 import sys
 import random
+import glob
 from wtcClient import wtcClient
 from JIRAClient import JIRAClient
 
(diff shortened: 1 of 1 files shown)


vlimant:
does not look like moduleLock can hold a lock https://github.com/CMSCompOps/WmAgentScripts/blob/master/utils.py#L1427 releases it for deadlocks.
there's likely a way to do things better

haozturk:
I could not understand the issue that this PR is trying to solve. I will close it. Feel free to re-open it if you find it necessary.

vlimant:
if actor fails somewhere, it might lead to hundreds of fictuous ACDC as mentioned in #528

PR #531, opened 2020-04-15
Feature disable nanoaod

### LINKED ISSUE #530: Exclude NANOD from Unified Output data placement
**Impact of the new feature**
Unified, WMCore 

We need to disable Output data placement from Unified for NANOAOD and NANOAODSIM. For more details, please see the relevant issue in WMCore:

https://github.com/dmwm/WMCore/issues/9634

Just disable it from Unified config for standard Workflows and DDM subscriptions (parameter: tier_to_DDM). In case it is needed to be done for RelVal too, we should do it either directly in the code or make use of the same parameter, even though for Relval output data placement it seems Unified is using Phedex instead DDM. 

N/A

**Additional context**
N/A

todor-ivanov:
Fixes #530 

#### Status
<not-tested>

#### Description
After our previous investigation of the code we concluded that Unified is doing output data placement through DDM for standard Workflows and through Phedex for RelVal. It seems we can benefit from the already existing parameter 'tier_to_DDM' here. And we can hard code the exclusion of NANOD for Relval in a similar way as it is done for the rest of the data tiers here:

https://github.com/CMSCompOps/WmAgentScripts/blob/master/Unified/closor.py#L502-L518

<YES>

#### Related PRs
NO
We may want to make it configurable for RelVal workflows too. 

@sharad1126 @amaltaro @vlimant

### DIFF
--- a/CondorMonitoring/JobCounter.py
+++ b/CondorMonitoring/JobCounter.py
@@ -46,10 +46,10 @@ overview_workflows = {}
 json_name_workflows = "CondorJobs_Workflows.json" # Output json file name
 
 ##SSB links
-site_link = "http://dashb-ssb.cern.ch/dashboard/templates/sitePendingRunningJobs.html?site="
-overalls_link = "http://dashb-ssb-dev.cern.ch/dashboard/templates/sitePendingRunningJobs.html?site=All%20"
-url_site_status = 'http://dashb-ssb.cern.ch/dashboard/request.py/getplotdata?columnid=158&batch=1&lastdata=1'
-url_site_pledges = 'http://dashb-ssb.cern.ch/dashboard/request.py/getplotdata?columnid=159&batch=1&lastdata=1'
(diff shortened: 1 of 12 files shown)


todor-ivanov:
@sharad1126 Once I made my PR to the testing branch, I saw something like 31 commits dragged from my repository together with my single line of change. I am committing from a  feature branch which in sync with 'upstream master'. The situation tells me that the current version of the 'testing' branch in this repository is not sync with master. Would you be able to take care of that? Then all that garbage from my PR should vanish.  It should not screw other's people work in testbed though. Thanks in advance.

amaltaro:
Todor, I'd suggest to make this PR against master. I don't think the "testing" branch is working as expected.

sharad1126:
@todor-ivanov I think you need to remove this PR and make it against the master branch as Alan suggested. I am actually planning to remove the testing branch and make a new one. It has come to my understanding that we need a much better approach for the testing branch.

PR #533, opened 2020-04-17
temparorily take out PIC from the distribution for tape storage

z4027163:
As requested by L2, PIC needs to be taken out of tape writing, temparorily set "usable" of PIC tape to be 0.

### DIFF
--- a/utils.py
+++ b/utils.py
@@ -2243,7 +2243,11 @@ class siteInfo:
 
             if mss in sites_space_override:
                 self.storage[mss] = sites_space_override[mss]
-
+            
+            ## disbale PIC tape temparorily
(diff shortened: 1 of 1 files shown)


vlimant:
the right place to set this is in 

http://cmsmonitoring.web.cern.ch/cmsmonitoring/storageoverview/latest/StorageOverview.json

used at 

https://github.com/CMSCompOps/WmAgentScripts/blob/master/utils.py#L2231

sharad1126:
@z4027163 , Jean-Roch is right. we should not put code in the system when there is a proper way of doing it specially when we use mss_usage which is a json coming from the cms monitoring url. 
https://github.com/CMSCompOps/WmAgentScripts/blob/master/utils.py#L1855

Issue #538, opened 2020-04-17
check dataset size for all output going to tape


vlimant:
how come that 1.2PB dataset made it to an tape request while the limit (way too high already) is 850TB ?

https://github.com/CMSCompOps/WmAgentScripts/blob/master/unifiedConfiguration.json#L217

it's because the check ins only done for the dataset passed to ddm

https://github.com/CMSCompOps/WmAgentScripts/blob/master/Unified/checkor.py#L1037

to be fixed

vlimant:
is that not an issue anymore ?

sharad1126:
I don't think there is any issues for the tape as such right now. I only see issues with disks at some sites. @z4027163 can you check and clarify this?

vlimant:
is there no interest to merge this protection for checking all relevant output sizes?

haozturk:
Unified does not deal with output data placement anymore, so closing this issue.

PR #539, opened 2020-04-18
Mss size check

vlimant:
solves #538 and gets further with checking independently the size of each dataset, not only the sum

### DIFF
--- a/Unified/checkor.py
+++ b/Unified/checkor.py
@@ -1223,6 +1223,14 @@ class CheckBuster(threading.Thread):
                 assistance_tags.add('bigoutput')
                 custodial = None
 
+            if custodial:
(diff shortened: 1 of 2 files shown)


sharad1126:
@vlimant why are we changing the size from 850TB to 300TB?

vlimant:
to let operation control when a very large output is out there. 300TB is already large.
you can relax it one #541 is in

sharad1126:
@z4027163  can you please cross check this and then merge it?

z4027163:
@vlimant The change looks fine.  But is the loop in L1221-1224 redundant? As it was covered in the loop at L1226.

vlimant:
L1226 (single output size check) does indeed cover L1221 (total output size check)

PR #544, opened 2020-04-22
caching repetitive calls to dbs

vlimant:
further on #462 #476 
affecting "only" showError.py I believe ( @z4027163 you had a recent look I think )


vlimant:
@z4027163 I think 6be97521fa242a24669127f707dad1aa810576c6 is introducing a missing "]"

I meant 

`    return [df['block_name'] for df in r][0] if r else None`

z4027163:
Sorry, my bad, changed it back

PR #563, opened 2020-05-19
stop failing on wmstats

vlimant:
would avoid GQ module to fail every other hour because of wmstats
FYI @amaltaro

### DIFF
--- a/utils.py
+++ b/utils.py
@@ -6818,7 +6818,7 @@ class workflowInfo:
 
             trials-=1
             time.sleep(1)
-        return None
+        return {}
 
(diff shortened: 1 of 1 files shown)


amaltaro:
Given that we are moving all the data location/placement away of unified, we can likely deprecate GQ as well (unless it does more than what I infer from its name).

vlimant:
there seem to be a monitoring for stuck WQE now in reqmgr ; good.
what is taking care of replacing the data automatically if necessary?
you can throw GQ in front of the bus, at least it will stop failing.

Issue #511, opened 2020-03-25
Removing Step chain message from Critical log


scarletnorberg:
It just causes extra messages that are not needed and not critical in the critical message log.

This is printed in critical logs: "the conversion of WF name to stepchain encoutered different value of Multicore 8 != 1"

Apparently it is something coming from the code that transform task chain to step chain. It’s not really an error.
see this log from the workflow you sent
invalidating the workflow and outputs by unified operator cmsunified, reason: Transform to StepChain
the conversion to stepchain encoutered different value of Multicore 8 != 1
Cloned into WF name by unified operator , reason: Transform to StepChain

PR #571, opened 2020-05-29
remove critical message about different core in stepchain conversion

z4027163:
Fixes #<GH_Issue_Number>
#511
#### Status
<In development | not-tested | on hold | ready>
N/A
#### Description
Remove the critical log when the different core is found in stepchain conversion, since different core  task chain can be converted as long as its production efficiency passes the threshold
<YES | NO>
YES

<Always mention people to check and review proposed changed.>
@sharad1126


sharad1126:
@z4027163 i think we can remove this and merge the PR as there is nothing we can do much if this message comes in critical logs.

PR #581, opened 2020-06-09
change eos mkdir

z4027163:
change mkdir in eos as suggested by eos team.
@sharad1126

### DIFF
--- a/cWrap.sh
+++ b/cWrap.sh
@@ -10,7 +10,7 @@ fi
 cd $BASE_DIR
 
 modulename=`echo $1 | sed 's/\.py//' | sed 's/Unified\///'`
-eos root://eoscms.cern.ch mkdir -p $HTML_DIR/logs/$modulename/
+mkdir -p $HTML_DIR/logs/$modulename/
 eos root://eoscms.cern.ch mkdir -p $FINAL_HTML_DIR/logs/$modulename/
 
 last_log=$HTML_DIR/logs/$modulename/last.log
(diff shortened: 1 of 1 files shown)


sharad1126:
@z4027163 the eos directory is `FINAL_HTML_DIR` only so let's change all the `cp` to `eos cp` and `mkdir` to the one you did but only for the ones where eos is concerned.

cWrap.sh
```diff
@@ -10,8 +10,8 @@ fi
 cd $BASE_DIR
 
 modulename=`echo $1 | sed 's/\.py//' | sed 's/Unified\///'`
-mkdir -p $HTML_DIR/logs/$modulename/
-mkdir -p $FINAL_HTML_DIR/logs/$modulename/
+eos root://eoscms.cern.ch mkdir -p $HTML_DIR/logs/$modulename/
```
sharad1126, line 13: this shouldn't be eos as HTML_DIR is not eos directory!

sharad1126:
@z4027163 can you please also check the `cp` to `eosp` in cWrap.sh script too wherever eos directory is used. it could help as well.

z4027163:
@sharad1126 done

sharad1126:
@z4027163  if we are sure about this PR that you added all eos infront of eos related cp and mkdir only, let's please merge this.

sharad1126:
merging the PR. the shell check fails because of double quotes which will be resolved in #594

PR #586, opened 2020-06-12
Set up flake8 test for github actions

ishanrai05:
#### Status
Tested on my branch

#### Description
Set up flake8 tests to check python syntax

yes

#### Related PRs
None

No.


.github/workflows/check_syntax.yml
```diff
@@ -6,5 +6,19 @@ jobs:
...
-      uses: andymckay/pycodestyle-action@0.1.3
+    - name: Checkout
+      uses: actions/checkout@v2.2.0
+    - name: Setup Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: 2.7
+    - name: Install Flake
+      run: |
+        python3 -m pip install flake8
```
sharad1126, line 17: should be python2

sharad1126:
@ishanrai05 if I am not wrong, this GH action is suppose to check the python syntax for the whole GH repo codebase right?
If that is the case, doing it every 6 hours won't make sense.

ishanrai05:
Should I change it to once a day? Or a week?

sharad1126:
yeah let's do it for now once a day and then we will do our best to fix the errors asap and then change the cronjob later to once a week or something(maybe)

Issue #588, opened 2020-06-17
checkor reported unsupported operand issue


sharad1126:
**Impact of the bug**
checkor module where we want to check for nanoaod tiers failed on a specific workflow. Looks like the total filecount returned by the function was `None`.

**Describe the bug**
```
failed on pdmvserv_task_B2G-RunIISummer16NanoAODv7-02956__v1_T_200609_205933_7388 due to unsupported operand type(s) for +: 'int' and 'NoneType' and Traceback (most recent call last):
File "Unified/checkor.py", line 374, in run
self.check()
File "Unified/checkor.py", line 1041, in check
phedex_presence[output] = sum(map(lambda x: x[1], all_filecount_pb))
TypeError: unsupported operand type(s) for +: 'int' and 'NoneType'
```

**How to reproduce it**
https://github.com/CMSCompOps/WmAgentScripts/blob/master/Unified/checkor.py#L1041
wf: http://dmytro.web.cern.ch/dmytro/cmsprodmon/workflows.php?prep_id=task_B2G-RunIISummer16NanoAODv7-02956

**Expected behavior**
should not report this error i.e. all_filecount_pb should not be `None`.

**Additional context and error message**
@amaltaro  @todor-ivanov looks like we encountered another new bug for nanoaod tier in checkor.

sharad1126:
actually looks like more than one failed:
```
failed on pdmvserv_task_B2G-RunIISummer16NanoAODv7-02487__v1_T_200428_100049_5406 due to unsupported operand type(s) for +: 'int' and 'NoneType' and Traceback (most recent call last):
File "Unified/checkor.py", line 374, in run
self.check()
File "Unified/checkor.py", line 1041, in check
phedex_presence[output] = sum(map(lambda x: x[1], all_filecount_pb))
TypeError: unsupported operand type(s) for +: 'int' and 'NoneType'
```

sharad1126:
#590 should be able to fix this. we are waiting to close this issue once we are satisfied.

todor-ivanov:
The PR #590  provided by @vlimant  should fix this. Thanks!

PR #590, opened 2020-06-18
Invalid rucio block lenght

vlimant:
"fixes" #588 

rucio returned 

> (u'/Wprimetotb_M2000W600_LH_TuneCP5_13TeV-madgraph-pythia8/RunIISummer16NanoAODv7-PUMoriond17_Nano02Apr2020_102X_mcRun2_asymptotic_v8-v1/NANOAODSIM#04ef69e5-9084-4649-89d7-8ce9552f786f', None)

for the block length.

BTW: rucio setup could be more integrated

### DIFF
--- a/RucioClient.py
+++ b/RucioClient.py
@@ -91,8 +91,10 @@ class RucioClient(Client):
             blocks = []
             for block in self.getBlockNamesDataset(dataset):
                 blocks.append((block, self.getFileCountBlock(block)))
+            if blocks is None:
+                raise Exception("number of file count per block is None")
         except Exception as e:
             print(str(e))
-            return []
+            return 0
         return blocks
 
--- a/RucioClient.py
+++ b/RucioClient.py
@@ -91,10 +91,8 @@ class RucioClient(Client):
             blocks = []
             for block in self.getBlockNamesDataset(dataset):
                 blocks.append((block, self.getFileCountBlock(block)))
-            if blocks is None:
-                raise Exception("number of file count per block is None")
         except Exception as e:
             print(str(e))
-            return 0
+            return []
         return blocks
 
--- a/RucioClient.py
+++ b/RucioClient.py
@@ -93,6 +93,6 @@ class RucioClient(Client):
                 blocks.append((block, self.getFileCountBlock(block)))
         except Exception as e:
             print(str(e))
-            return 0
+            return []
         return blocks
 
--- a/RucioClient.py
+++ b/RucioClient.py
@@ -75,6 +75,8 @@ class RucioClient(Client):
         """
         try:
             numFiles = self.get_metadata(self.scope, block)['length']
+            if numFiles is None:
+                raise Exception("block length in rucio is None")
         except Exception as e:
             print(str(e))
             return 0


sharad1126:
thanks @vlimant . that's a good catch as I was expecting something similar but it's not coming from the function you updated here instead it is coming from `getFileCountPerBlock`.

RucioClient.py
```diff
@@ -91,8 +91,10 @@ def getFileCountPerBlock(self, dataset):
             blocks = []
             for block in self.getBlockNamesDataset(dataset):
                 blocks.append((block, self.getFileCountBlock(block)))
+            if blocks is None:
+                raise Exception("number of file count per block is None")
         except Exception as e:
             print(str(e))
-            return []
+            return 0
```
vlimant, line 98: the function getFileCountPerBlock needs to return a list, empty if there's an exception

RucioClient.py
```diff
@@ -91,8 +91,10 @@ def getFileCountPerBlock(self, dataset):
             blocks = []
             for block in self.getBlockNamesDataset(dataset):
                 blocks.append((block, self.getFileCountBlock(block)))
+            if blocks is None:
```
vlimant, line 94: blocks=[] with subsequent append cannot turn into None, ever. 
are you actually observing this ?

vlimant:
@sharad1126 have you tried to run checkor on task_B2G-RunIISummer16NanoAODv7-02956 with the changes I commited ? I cannot (see #591)

RucioClient.py
sharad1126: If you see the GH issue, checkor reports issue on L1041 where the None is coming from ‘all_filecount_pb’ and that is the sum of ‘getfilecountperblock’ for Phedex and rucio! I don’t see how changing ‘getfilecountblock’ would help here as that function isn’t being called!

sharad1126:
@vlimant how do you run checkor on a specific wf? I have never seen any option like -w or something which restricts checkor to run on any particular wf

sharad1126:
@vlimant also the function https://github.com/CMSCompOps/WmAgentScripts/blob/master/RucioClient.py#L83 which is called in https://github.com/CMSCompOps/WmAgentScripts/blob/master/Unified/checkor.py#L1031 takes input as dataset and your comment where you tried it for block. @todor-ivanov see I told you we should not put in functions we won't need for now as it creates confusions.

RucioClient.py
sharad1126: the function returns the number of files not a list as the description explains below the function. https://github.com/CMSCompOps/WmAgentScripts/blob/master/RucioClient.py#L74
https://github.com/CMSCompOps/WmAgentScripts/blob/master/RucioClient.py#L85

sharad1126: I see your point. my bad

RucioClient.py
todor-ivanov: @vlimant for fixing `getFileCountBlock()`.  I simply did not expect to see a `None` value returned by Rucio here. BTW @sharad1126 this function is called and it is exactly here: 
https://github.com/CMSCompOps/WmAgentScripts/blob/1eb640459848d65de6188d44e5a42fe789098091/RucioClient.py#L93
And this is exactly how the `None` value enters the final block list. This `None` is not returned as  the full block list but it rather enters as one of it's elements.

And this return a `0` instead of `[]` in `getFileCountPerBlock()`, somehow slipped between the lines. Thanks for catching that  one too.

vlimant:
could @amaltaro check quikly how a None was maybe injected in Rucio in the first place?
otherwise one has to create an issue to rucio jira (@ericvaandering)

ericvaandering:
I was worried that this was just a metadata field which could be out of sync with list-contents, but I don't think so: https://github.com/rucio/rucio/blob/06b662cd4d0561861c8c23759e83eb6715f28f30/lib/rucio/core/did.py#L1149

Anyhow, None is the intended "length" for empty collections

PR #591, opened 2020-06-18
central rucio config

vlimant:
Allows a user to run rucio client (while fixing #590) , and unified modules as oneself (as should be BTW)

### DIFF
--- a/RucioClient.py
+++ b/RucioClient.py
@@ -9,7 +9,10 @@ Environment:
     ${RUCIO_HOME}/rucio.cfg
 """
 
+import os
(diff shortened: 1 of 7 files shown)


sharad1126:
The way it is right now also enables the user to run the rucio client if one has the rucio.cfg file. So I don't think doing this for the production code base is necessary.

vlimant:
having rucio.cfg in the repo and used, allows for controlling exactly what goes in it, and not depend on what a P&R would have in his.
on top of that 932345a9a719aa421b358192ed8654b881818696 is necessary, otherwise rucio complains that user X cannot use account "unified"

PR #600, opened 2020-06-29
run pylint on pull requests

ishanrai05:
#### Status
Ready

#### Description
Run flake8 and pylint tests on git diff for every PR

Yes

#### Related PRs
None

None

### DIFF
--- a/.github/workflows/python_lint.yml
+++ b/.github/workflows/python_lint.yml
@@ -1,0 +1,25 @@
+name: PyLint test
+
+on: [push, pull_request]
+
+jobs:
+  lint-python:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2.2.0
+    - name: Get Diff Action
+      uses: technote-space/get-diff-action@v1.2.8
+      with:
+        SUFFIX_FILTER: .py
+    - name: Setup Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: 2.7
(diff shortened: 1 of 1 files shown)


.github/workflows/python_lint.yml
```diff
@@ -0,0 +1,32 @@
...
+      uses: actions/setup-python@v2
+      with:
+        python-version: 2.7
+    - name: Install Flake8 and Pylint
+      run: |
+        python2 -m pip install flake8 pylint
+    - name: Lint with flake8
+      run: |
+        # stop the build if there are Python syntax errors or undefined names
+        flake8 ${{ env.GIT_DIFF }} --count --select=E9,F63,F7,F82 --show-source --statistics
```
ishanrai05, line 25: * E9 tests are about Python syntax errors usually raised because flake8 can not build an Abstract Syntax Tree (AST). Often these issues are a sign of unused code or code that has not been ported to Python 3. These would be compile-time errors in a compiled language but in a dynamic language like Python they result in the script halting/crashing on the user.
* F63 tests are usually about the confusion between identity and equality in Python. Use ==/!= to compare str, bytes, and int literals is the classic case. These are areas where a == b is True but a is b is False (or vice versa).
* F7 tests logic errors and syntax errors in type hints
* F82 tests are almost always undefined names which are usually a sign of a typo, missing imports, or code that has not been ported to Python 3. These also would be compile-time errors in a compiled language but in Python a NameError is raised which will halt/crash the script on the user.

PR #604, opened 2020-07-07
Add alternative pilot from SubRequestType

### LINKED ISSUE #585: Alternative way to tag pilot workflows besides "pilot" processing string
**Impact of the new feature**
WMAgent? 

Only way to indicate that workflow is a pilot is to add "pilot" processing string. Processing string is always added to output dataset name. In some cases this is intended - to show that this dataset is not suitable for analysts, however, in other cases output is suitable for analysis and even though it is a pilot workflow, it would be better to *not to* include it in output in order to not confuse analysts.

We would like to ask for a way to indicate that workflow is pilot without adding "pilot" as processing string. We would like to add it as processing string only in some cases, but also have a way to indicate it without reflecting that in output dataset name. It could be some parameter in "Comments" field (to be discussed with ReqMgr2 team).

It could also be any other argument, such as SubRequestType, but this would require deeper investigation and changes in McM as SubRequestType is already used in some cases.

**Additional context**
We use ReqMgr2 to submit workflows.

A couple of emails from @srimanob in our email thread:
```
We don't propose to delete what we are using now, we propose to have an alternative
way.

+ Pilot as processing string is useful, i.e. for pilot requests at the beginning of
  campaign. We don't expect analyst to use them as it runs quick, no approval from GEN
  conveners. The goal is to test the overall production.

+ Pilot as a comment is an additional option we would like to have. This is useful for
  validating reMiniAOD or reNanoAOD campaign, where we start from PROD AOD or MiniAOD.
  The output datasets are useful for analyst. So we would like to avoid having "pilot"
  as a processing string. However, we would like to short the communication process to
  assign these workflows manually during pilot period also.
```
```
In short, as of now, the "pilot" in processing string serves 2 purposes:
+ to trigger the workflow as pilot ==> To avoid delay of manual assigning workflow
+ to add to processing string

I would like to have the way to trigger that workflow is pilot without necessary of
adding "pilot" as processing string. We will have it on what we want only. "pilot" as
processing string is useful in some cases, but should be avoided in others.

Useful case:
We open a new GS, DR campaign, the first request we put in should have the word
"pilot" somewhere.
/TTJets_14TeV/Run3Summer22GS-pilot-xxxx-v1/AOD => useful to have "pilot" there, so
analyst will not use it

Case that I would like to avoid to have pilot in the processing string:
ReMiniAOD or ReNanoAOD because input datasets that we uses are big prod ones like
TTJets_Hadronic, or some Higgs process. The output is useful for analyst.
```

z4027163:
Fixes #585 

#### Description
An alternative way to tag pilot workflows besides "pilot" processing string, in order to have the normal output dataset name.

@sharad1126 @amaltaro

### DIFF
--- a/utils.py
+++ b/utils.py
@@ -8086,7 +8086,10 @@ class workflowInfo:
 		else:
 		    print(msg)
 		return True
-	
+        # If there is 'pilot' in SubRequestType (an alternative pilot)
+        if 'SubRequestType' in self.request and 'pilot' in self.request['SubRequestType']:
+            return True        
+ 	
         for campaign,label in pas:
             if not CI.go( campaign, label):
                 if log:
--- a/utils.py
+++ b/utils.py
@@ -8088,6 +8088,7 @@ class workflowInfo:
 		return True
         # If there is 'pilot' in SubRequestType (an alternative pilot)
         if 'SubRequestType' in self.request and 'pilot' in self.request['SubRequestType']:
+            print "Alternative pilot"
             return True        
  	
         for campaign,label in pas:


sharad1126:
@z4027163 I see there are two `def go` and you added this to only one function. are we sure the other one isn't used?

z4027163:
@sharad1126 I think the pilot check in the "def go" of campaignInfo is redundant. The function is only used in the [1], and the workflowinfo.go function at [1] will return "True" before this line if it's a pilot.

[1]https://github.com/CMSCompOps/WmAgentScripts/blob/28e202de07675ee909a90a4824cbad1ada6cb528/utils.py#L8091

sharad1126:
@z4027163 I think we should checkout this PR. We can ask PPD to inject a test pilot workflow.

sharad1126:
@z4027163 I think we can merge this PR and then ask PPD to send the next `pilot` with the `pilot` keyword in the `subrequest` and see how it goes. what do you think?

srimanob:
Should we submit a test?

z4027163:
@srimanob yes, please, we are about to ask for it.

sharad1126:
@srimanob I have just sent an email to request the same. thanks

PR #609, opened 2020-07-13
add GEN-SIM-DIGI-RAW-MINIAOD data tier to_DDM

sharad1126:
Adding GEN-SIM-DIGI-RAW-MINIAOD to unified configuration

#### Status
ready

#### Description
Phat and PdMv has requested to add a new data tier - GEN-SIM-DIGI-RAW-MINIAOD and as it would be an output dataset, the tier needs to go to DDM.

yes

#### Related PRs
N/A

DBS

@amaltaro @z4027163 @todor-ivanov fyi

### DIFF
--- a/unifiedConfiguration.json
+++ b/unifiedConfiguration.json
@@ -227,7 +227,7 @@
    "description": "The data tiers that do not go to AnaOps"
  },
  "tiers_to_DDM": {
-   "value" : ["AODSIM","MINIAODSIM","GEN-SIM-RAW","GEN-SIM-RECO","GEN-SIM-RECODEBUG","AOD","RECO","MINIAOD","ALCARECO","USER","RAW-RECO","RAWAODSIM","NANOAOD","NANOAODSIM","FEVT","PREMIX"],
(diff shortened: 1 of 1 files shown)


amaltaro:
@todor-ivanov do we need to change anything in the MSOutput?

sharad1126:
@amaltaro most likely @todor-ivanov would have to only add the data tier where you would have added the others going to DDM.

todor-ivanov:
@amaltaro @sharad1126 I think nothing needs to be done in MSOutput regarding this new data tier. From what I see in the `checkor.py` module the `tiers_to_ddm` parameter at the end is used only for custodial outputs: [1]. And we do not deal with them in MSOutput. In our case DDM is the default data management system so if this data tier happens to be a valid output dataset it will go to DDM,  and if there is no any new (specific to the new data tier) condition regarding RelVal outputs then we do not have to do anything in MSOutput. 


[1]
https://github.com/CMSCompOps/WmAgentScripts/blob/229ca946a9d461b8d80e1aec8eec604fad650533/Unified/checkor.py#L1264

PR #612, opened 2020-07-15
change eos I/O command

### LINKED ISSUE #611: using eos cp everywhere in the code instead of cp
**Impact of the new feature**
we have eos fuse and fusex in our machines which is causing issues while cp, ls, mkdir, and read. Maria from eos team has suggested to do all use fuse so we use eos cp, eos mkdir, eos ls, etc. For reading a file, the option is to copy it in cache and then read it from there. 

The modules get stuck from time to time with the uninterrupted sleep due to I/O issues which is most likely eos. https://github.com/CMSCompOps/WmAgentScripts/blob/master/utils.py#L1208 should be eos cp instead of cp. All other places where we are doing cp or mkdir or ls for eos directory, we should use eos before the command with the env variable like in https://github.com/CMSCompOps/WmAgentScripts/pull/605/files

Use similar https://github.com/CMSCompOps/WmAgentScripts/pull/605/files in whole code interacting with eos. 

nothing for now but remove eos dependency is a long term goal. 

**Additional context**
@z4027163 fyi

z4027163:
Fixes #611 
change eos interaction with eos command.

For reading/writing/accessing/moving files in eos, add eos ENV set up and eos command.

@sharad1126

### DIFF
--- a/Unified/htmlor.py
+++ b/Unified/htmlor.py
@@ -19,7 +19,7 @@ def htmlor( caller = ""):
 
     for backup in ['statuses.json','siteInfo.json','equalizor.json']:
         print "copying",backup,"to old location"
-        os.system('cp %s/%s /afs/cern.ch/user/c/cmst2/www/unified/.'%(monitor_pub_dir, backup))
+        os.system('env EOS_MGM_URL=root://eoscms.cern.ch eos cp %s/%s /afs/cern.ch/user/c/cmst2/www/unified/.'%(monitor_pub_dir, backup))
         #os.system('cp %s/%s %s/.'%(monitor_dir, backup, monitor_pub_dir))
 
     try:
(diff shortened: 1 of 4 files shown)


sharad1126:
@z4027163 I hope you have made sure that all cp and mkdir you replaces with eos cp and eos mkdir are actually copying either from/to eos and similarly making directory in eos using mkdir. I am not sure if eos rm and eos mv is something that works, so I will confirm this with the eos team. I guess it should work though. 
If you see the GH issue, there is no eos function called eosFile or eosRead. They are inbuilt unified functions which also need changes i.e. either use xrootd python bindings to directly read from eos or just eos cp into cache first and then read directly.

https://github.com/CMSCompOps/WmAgentScripts/blob/master/utils.py#L1208 needs to be changed to eos cp.

sharad1126:
ok I confirmed with eos team that eos rm and eos mv is good to use. @z4027163 fyi

Unified/htmlor.py
```diff
@@ -19,7 +19,7 @@ def htmlor( caller = ""):
 
     for backup in ['statuses.json','siteInfo.json','equalizor.json']:
         print "copying",backup,"to old location"
-        os.system('cp %s/%s /afs/cern.ch/user/c/cmst2/www/unified/.'%(monitor_pub_dir, backup))
+        os.system('env EOS_MGM_URL=root://eoscms.cern.ch eos cp %s/%s /afs/cern.ch/user/c/cmst2/www/unified/.'%(monitor_pub_dir, backup))
```
sharad1126, line 22: I don't think this is eos cp as it's doing cp in afs and not eos directory.

Unified/htmlor.py
```diff
@@ -1553,8 +1553,8 @@ def site_div_header(desc):
 
 
     eosFile(time.strftime("%s/summary_%%Y_"%(monitor_dir), time.gmtime())+this_week+".json").write(json.dumps(summary_content, indent=2)).close()
-    os.system(time.strftime("cp %s/summary_%%Y_"%(monitor_dir), time.gmtime())+this_week+".json %s/summary.txt"%(monitor_pub_dir))
-    os.system(time.strftime("cp %s/summary_%%Y_"%(monitor_dir), time.gmtime())+last_week+".json %s/last_summary.txt"%(monitor_pub_dir))
+    os.system(time.strftime("env EOS_MGM_URL=root://eoscms.cern.ch eos cp %s/summary_%%Y_"%(monitor_dir), time.gmtime())+this_week+".json %s/summary.txt"%(monitor_pub_dir))
+    os.system(time.strftime("env EOS_MGM_URL=root://eoscms.cern.ch eos cp %s/summary_%%Y_"%(monitor_dir), time.gmtime())+last_week+".json %s/last_summary.txt"%(monitor_pub_dir))
```
sharad1126, line 1557: same here. these are also afs copy and not eos copy.

Unified/htmlor.py
z4027163: https://github.com/CMSCompOps/WmAgentScripts/blob/master/utils.py#L97
I see it's actually using eos here.

PR #616, opened 2020-07-17
query once the size of output to be checked

vlimant:
sorry for leaving untested code in #539.
this is slightly better than #614 @z4027163 IMO


vlimant:
isn't that more economical of DBS calls @z4027163 ?

PR #617, opened 2020-07-24
add a serviceconfig file to change endpoint urls

sharad1126:
allows to run tests in testing machine using the testbed endpoints

#### Status
ready

#### Description
The urls in utils.py are imported from serviceconfig json file which allows to have production endpoints in master branch and testbed urls in testing branch

no

#### Related PRs
n/a

none 

@z4027163

### DIFF
--- a/serviceConfiguration.json
+++ b/serviceConfiguration.json
@@ -1,0 +1,62 @@
+{
+    "mongo_db_url":  {
+        "value": "vocms0274.cern.ch",
+        "description": "The mongodb instance."
+    },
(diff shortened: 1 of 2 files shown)


sharad1126:
@ishanrai05 can you have a look here? the git diff isn't working as it shows the warnings and errors in the whole python script. Also, it's showing some errors which I don't think are errors like `utils.py:  4:0: [E0401] Unable to import 'dbs.apis.dbsClient' (import-error)` as it is imported properly without issues.

PR #625, opened 2020-07-31
remove noncustodial site initially

### LINKED ISSUE #622: T1 sites are chosen to be the output site despite out of space
The unified keeps sending WFs to T1 sites (PIC, KIT) despite they have 0 space available. This causes the long time issue that the data are not evenly distributed among the sites.

**Describe the bug**
In the assignor where it decides the output site [1]. If any T1 sites are in the "sites_allowed",  only T1 sites will be considered even they are out of space. If all the T1 sites are out of space, the function [2] will pick the first site in the array, and it ends up with "PIC" usually.

Is picking T1 sites only a policy? I would suggest at least check the remaining quota as well at [3]. 

[1] https://github.com/CMSCompOps/WmAgentScripts/blob/master/Unified/assignor.py#L298-L304
[2] https://github.com/CMSCompOps/WmAgentScripts/blob/master/utils.py#L2565
[3] https://github.com/CMSCompOps/WmAgentScripts/blob/master/Unified/assignor.py#L298

@sharad1126  @vlimant  Do you know the reason why it was set in this way?

z4027163:
Fixes #622 

#### Description
Remove the noncustodial site assignment, leaving it empty initially.
@sharad1126 @vlimant

### DIFF
--- a/Unified/assignor.py
+++ b/Unified/assignor.py
@@ -307,8 +307,8 @@ def assignor(url ,specific = None, talk=True, options=None):
         wfh.sendLog('assignor',"Placing the output on %s"%sites_out)
         parameters={
             'SiteWhitelist' : sites_allowed,
-            'NonCustodialSites' : sites_out,
-            'AutoApproveSubscriptionSites' : list(set(sites_out)),
+            'NonCustodialSites' : [],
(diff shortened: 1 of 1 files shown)


sharad1126:
@z4027163  should be close this PR?

PR #627, opened 2020-08-04
showError tar is causing I/O issues as it uses eos

### LINKED ISSUE #611: using eos cp everywhere in the code instead of cp

sharad1126:
Fixes #611 

#### Status
not-tested

#### Description
I saw today some machines were having D state for the tar processes as it also communicates with eos. We might want to cross check again all scripts using eos.

yes

#### Related PRs
#579 

eos

@z4027163 @haozturk fyi

### DIFF
--- a/Unified/showError.py
+++ b/Unified/showError.py
@@ -117,7 +117,7 @@ class LogBuster(threading.Thread):
 
                                     if os.system('ls %s'% local)==0:
                                         os.system('env EOS_MGM_URL=root://eoscms.cern.ch eos mkdir -p %s'%(m_dir))
-                                        os.system('tar zxvf %s -C %s'%(local,m_dir))
+                                        os.system('env EOS_MGM_URL=root://eoscms.cern.ch eos tar zxvf %s -C %s'%(local,m_dir))
                                         ## truncate the content ??
                                         actual_logs = os.popen('find %s -type f'%(m_dir)).read().split('\n')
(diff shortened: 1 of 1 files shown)


z4027163:
eos doesn't have the tar command currently. Does "tar fist at /tmp and eos mv" work?

sharad1126:
yeah I tried and looks like it doesn't works. The solution in this case I think would be to copy the file to cache(https://github.com/CMSCompOps/WmAgentScripts/blob/master/utils.py#L105) and then tar it and then copy it back.

PR #630, opened 2020-08-05
Change touch and tar command on eos to avoid IO issue

z4027163:
Add eos command for touch and tar.

For tar command, tar the file at /tmp and then use "eos cp" command to move to eos. Since "eos mv" command doesn't support overwrite option (-T). 

@sharad1126 @haozturk


sharad1126:
Looks good to me.

PR #632, opened 2020-08-06
remove T1 only contraints for output placement.

z4027163:
For the output data placement noncustodial site selection, T1 only constraint is no longer needed. As discussed with L2, the site for output data placement is selected among T1+T2 sites based on their status and size. 

@sharad1126 @haozturk

### DIFF
--- a/Unified/assignor.py
+++ b/Unified/assignor.py
@@ -295,13 +295,12 @@ def assignor(url ,specific = None, talk=True, options=None):
             continue
 
  
-        t1_only = [ce for ce in sites_allowed if ce.startswith('T1')]
-        if t1_only:
+#        t1_only = [ce for ce in sites_allowed if ce.startswith('T1')]
+#        if t1_only:
             # try to pick from T1 only first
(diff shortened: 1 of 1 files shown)


sharad1126:
@z4027163  please make sure it doesn't picks up any T3 site.

sharad1126:
also, does this mean, https://github.com/CMSCompOps/WmAgentScripts/pull/625 is not needed?

z4027163:
@sharad1126 Most of the sites is T1+T2. The only exception is when the secondary is not heavy so that the sites include "add_on_good_aa " site (which are T3). But those settings where already used in the unified if no T1 sites are allowed. 

Do we want to run on T3 sites if it meets the requirement of AAA? Or else I can change them to run on T1 and T2 at any circumstances.

#625 is not needed.

sharad1126:
no T3 sites. only T1+T2 sites for sending the output to disks for noncustodial sites please.

z4027163:
Now changed to T1T2 only requirement.

PR #649, opened 2020-09-16
Skip string for rucio testing

z4027163:
This change is for rucio testing WF. Ignore WFs that contain string 'rucio' in several modules unless those are run manully. 
Therefore we can create a separate branch where we can run those module manually on rucio testing WF.


haozturk:
Changes in all modules except the checkor module look good to me.

PR #685, opened 2020-10-14
Opening the 20UL*Gen and the 20UL*pLHE campaigns that had pilot issue…

scarletnorberg:
"Opening the 20UL*Gen and the 20UL*pLHE campaigns that had pilot issues last week, and adding the 20UL18Minbais to the digi campaign."


haozturk:
@scarletnorberg Can you remove the space in the beginning of the parameter defined in line 2242?

PR #698, opened 2020-10-22
add minbias sitewhitelist intersection with available sites

### LINKED ISSUE #689: Workflows got stuck in assigned state due to a bug in assignor
**Impact of the bug**
Workflows are stuck in assigned if
1. It reads a PU
2. `TrustPUSitelists` is false
3. All site/sites hosting the PU is in drain

**Describe the bug**
When all the conditions described above is met, the workflows are stuck in assigned. Because `assignor` do not include the site/sites in drain in the sitewhitelist and even if the site stops draining at some point, the workflow cannot proceed, since the required site/sites is/are not in the sitewhitelist.

**How to reproduce it**
Original workflow of the following request is an example (CNAF hosts the PU and it was in drain during the assignment)
https://dmytro.web.cern.ch/dmytro/cmsprodmon/workflows.php?prep_id=task_PPD-RunIISummer20ULPrePremix-00004

**Expected behavior**
`Assignor` should keep the request in `assignment-approved` and it should make the assignment when the required site stops draining. In the current status, a manual abort/clone/assign operation is needed, which can be automated. 

**Additional context and error message**
@z4027163  @amaltaro  @todor-ivanov  FYI

z4027163:
Fixes #689 
Add a requirement of at least one of the minbias sitewhitelist in allowed sites. If no intersection is found, skip the workflow. 

The piece of code is tested.

@haozturk

### DIFF
--- a/Unified/assignor.py
+++ b/Unified/assignor.py
@@ -281,7 +281,19 @@ def assignor(url ,specific = None, talk=True, options=None):
         #        sendLog('assignor','%s has no whitelist'% wfo.name, level='critical')
         #    n_stalled+=1
         #    continue
-
+        
+        #check PU sitewhitelist overlap with sitewhitelist
+        minbias_site=[]
+        for campaign in wfh.getCampaigns():
+            if campaign in CI.campaigns and 'secondaries' in CI.campaigns[campaign]:
+                tem_sec=CI.campaigns[campaign]['secondaries']
+                for sec in secondary:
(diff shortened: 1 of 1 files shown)


haozturk:
@z4027163 in the last dev meeting, we decided to remove `SiteWhitelist` parameter for secondaries in the campaign configuration. But, this PR relies on it. We need to get the minbias location from somewhere else. Let me know if this does not make sense

z4027163:
@haozturk we can close this PR, the similar function is implemented already.

PR #711, opened 2020-11-11
Comment out some phedex function

z4027163:
Comment out some phedex function that are still used in the module.


haozturk:
None of these functions are used by any module right now. Instead of keeping the code, why don't we remove them?

z4027163:
Oh, you comment out the whole inputcycle. But remainor and equalizor are still running, and those use the function "getDatasetPresence" and "allCompleteToAnaOps"

haozturk:
They are not used in equalizor. Only use case is the remainor as you pointed out. What is the value of this module?

PR #740, opened 2020-12-09
Adding in new MiniAOD campaigns

scarletnorberg:
dealing with PRCAMPAIGNS-168 to PRCAMPAIGNS-178 where RunIISummer20UL*MiniAOD*v2 and the MiniAODUL*Data campaigns 

As well as PRCAMPAIGNS-158 PRCAMPAIGNS-165 which was removing the fractionpass all nanoAOD campaigns have fractionpass removed we will see how this goes.

### DIFF
--- a/campaigns.json
+++ b/campaigns.json
@@ -283,6 +283,24 @@
     }, 
     "resize": "auto", 
     "tune": true
+  },
+  "MiniAODUL16Data": {
+    "go": false,
+    "lumisize": -1,
+    "maxcopies": 1,
+    "resize": "auto"
+  },
+  "MiniAODUL17Data": {
+    "go": false,
+    "lumisize": -1,
+    "maxcopies": 1,
+    "resize": "auto"
+  },
+  "MiniAODUL18Data": {
+    "go": false,
+    "lumisize": -1,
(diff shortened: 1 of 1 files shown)


haozturk:
@scarletnorberg This PR has conflicts as you also can see. Most probably, it is because, you did not pull the latest changes before making your changes. My recommendation for avoiding this is the following:
```
git checkout master
git pull
git checkout -b <your_branch>
```

Note that this requires you to create new branches for campaigns, which I still argue that it is a better practice. 

For this PR specifically, can you fix the conflicts? Thanks

haozturk:
More dangerously, when you load these campaigns to database, you are not obeying the version in the master, which is the most up-to-date version. For this PR specifically, in your version, `fractionpass` is `.95` for some campaigns whereas it is `1` in master, which I did couple of days ago. By loading the campaigns.json file in this branch, you again changed `fractionpass` to `.95` in the database, which is a very bad thing and will affect production until we fix it.

haozturk:
I have fixed the conflicts. I kept the `"fractionpass": 1` for Data campaigns, which I think is safer than removing the parameter and falling back to the default=1 as it is more explicit.

PR #747, opened 2020-12-15
Opening campaigns that were successful with pilots and changing fract…

scarletnorberg:
MiniAODUL18Data PRCAMPAIGNS-169 pilot successful
MiniAODUL17Data PRCAMPAIGNS-168 pilot successful
RunIISummer19UL18MiniAODv2 PRCAMPAIGNS-174 pilot successful

Also updating all MC nanoAOD campaigns to have a fraction pass. Also updating the reminiAOD ones because assume that also needs to be higher then 95%. Data ones are left the same though.

### DIFF
--- a/campaigns.json
+++ b/campaigns.json
@@ -541,7 +541,7 @@
     "resize": "auto"
   }, 
   "Run3Summer19NanoAOD": {
-    "fractionpass": 0.985, 
+    "fractionpass": 0.99, 
     "go": true, 
     "lumisize": -1, 
     "maxcopies": 1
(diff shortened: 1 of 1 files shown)


haozturk:
@scarletnorberg Can we have a fractionpass of `.99` instead of `.985` as we agreed on the former value with PdmV/PPD. Probably, the latter is also okay, but, I would prefer to have the exact value in order to avoid confusions in the future.

scarletnorberg:
I could not remember what dima said and if 0.99 was still going to be more work. I wanted to set it to 0.98 I thought I would split the difference. I guess we can set it to 0.99 and if it is more work change it to something slightly lower which should be less work.

scarletnorberg:
Updated

scarletnorberg:
@haozturk updated

PR #791, opened 2021-02-19
Adding MinBais PU and two Snowmass Campaigns

scarletnorberg:
PRCAMPAIGNS-183 Run3Winter21DRMiniAOD adding PU 
PRCAMPAIGNS-187 SnowmassWinter21wmLHEGEN new campaign
PRCAMPAIGNS-185 SnowmassWinter21GEN new campaign

### DIFF
--- a/campaigns.json
+++ b/campaigns.json
@@ -256,8 +256,8 @@
   }, 
   "MiniAODHI18Data": {
     "SiteWhitelist": [
-     "T2_US_Vanderbilt"
-        ],
+      "T2_US_Vanderbilt"
+    ], 
     "fractionpass": 1.0, 
     "go": true, 
     "lumisize": -1, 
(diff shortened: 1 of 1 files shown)


haozturk:
@scarletnorberg  This PR removes the `pPb816Spring16pLHE` campaign. I suspect this is not intentional. If you agree, can you please fix it?
Another comment in general: When I see the diff in the PR, I see there are lots of additions and removals of the same lines. Since you remove and add the same line, it does not make a difference, but it makes the job of the reviewer harder. Could you please pay attention to it in the future PRs?
@z4027163  FYI

scarletnorberg:
How does this even make sense I did the PR for this https://github.com/CMSCompOps/WmAgentScripts/commit/b225953c9b1bec639045304099e3737f5e3f79b9
So it had to have been from update campaigns. I also pulled from master but this should have been merged into master already. 

It is also here: https://cmsweb.cern.ch/reqmgr2/data/campaignconfig/ALL_DOCS

Can you tell me where it went @haozturk ?

scarletnorberg:
It is not that I add or remove the same line. I ran load because I was worried about adding the new campaigns correctly. And when adding you run. a script and that script does apparently a shitty job.

PR #856, opened 2021-06-22
Adding some minbais's to various campaigns

scarletnorberg:
PRCAMPAIGNS-204
PRCAMPAIGNS-200

These two campaigns need three Minbais.


scarletnorberg:
PRCAMPAIGNS-208

Issue #855, opened 2021-06-21
[Py2Py3] Migrate helper functions talking to ReqMgr [1]


haozturk:
**Impact of the new feature**
Almost every unified module

**Is your feature request related to a problem? Please describe.**
As part of python3 migration, we should also migrate functions talking to ReqMgr in utils.py. I already migrated the following function which might be a guide for the further ones:
1. Old version: https://github.com/CMSCompOps/WmAgentScripts/blob/b34084adaa9d158bb4a2dd6a8dd16f448f0fba1a/utils.py#L2885 
2. New version: https://github.com/CMSCompOps/WmAgentScripts/blob/python3-migration/src/python/Services/ReqMgr/ReqMgrReader.py#L31 
3. Test: https://github.com/CMSCompOps/WmAgentScripts/blob/python3-migration/test/python/Services_t/ReqMgr_t/ReqMgrReader_t.py#L20

**Describe the solution you'd like**
Following functions looks like good candidates:
1. getWorkload: https://github.com/CMSCompOps/WmAgentScripts/blob/b34084adaa9d158bb4a2dd6a8dd16f448f0fba1a/utils.py#L2861
2. getWorkflowByOutput: https://github.com/CMSCompOps/WmAgentScripts/blob/b34084adaa9d158bb4a2dd6a8dd16f448f0fba1a/utils.py#L2901
3. getWorkflowById: https://github.com/CMSCompOps/WmAgentScripts/blob/b34084adaa9d158bb4a2dd6a8dd16f448f0fba1a/utils.py#L2950
4. getReqmgrInfo: https://github.com/CMSCompOps/WmAgentScripts/blob/b34084adaa9d158bb4a2dd6a8dd16f448f0fba1a/utils.py#L2871 
5. getAgentConfig: https://github.com/CMSCompOps/WmAgentScripts/blob/b34084adaa9d158bb4a2dd6a8dd16f448f0fba1a/utils.py#L3580
6. getWorkflowsByName: https://github.com/CMSCompOps/WmAgentScripts/blob/b34084adaa9d158bb4a2dd6a8dd16f448f0fba1a/utils.py#L3623
7. getWorkflows: https://github.com/CMSCompOps/WmAgentScripts/blob/b34084adaa9d158bb4a2dd6a8dd16f448f0fba1a/utils.py#L3649

There are other functions as well talking to ReqMgr, but let's work on them in the next issue. 

**Describe alternatives you've considered**
None

**Additional context**
[...]

PR #859, opened 2021-06-22
Migrate reqmgr functions to python3

amanda-grp:
#### Status
In development.
Passing all unit tests.

#### Description
Migration of helper functions talking to reqmgr in utils.py to python3, as described in #855:
- `getWorkload`
     => renamed to `getWorkflowSchema`
- `getWorkflowByOutput`
     => renamed to `getWorkflowsByOutput`
     => endpoint changed to RESTful API
- `getWorkflowById`
     => renamed to `getWorkflowsByPrepId`
     => endpoint changed to RESTful API
- `getReqmgrInfo`
- `getAgentConfig`
- `getWorkflowsByName`
     => renamed to `getWorkflowsByNames`
- `getWorkflows`
     => renamed to `getWorkflowsByStatus`
- `getSchema`
     => deprecated, use `getWorkflowSchema` instead
- `getSplittingsNew`
     => renamed to `getSplittingsSchema`

I also changed the `WebTools.getResponse` function so that it can handle params with list values and added the retry function to be used as a decorator in Utils/Decorators.py.

@haozturk


haozturk:
Thanks a lot Amanda, let me give you some data for testing. Note that we can use production for testing in this stage of the project as long as the API calls are lightweight. We can switch to testbed in a later stage. 

1. getWorkflow:
Check for https://cmsweb.cern.ch/reqmgr2/data/request/pdmvserv_task_BPH-RunIIFall18GS-00350__v1_T_201021_154340_8354 
This response should not change over time

2. getWorkflowByOutput:
Check for https://cmsweb.cern.ch/reqmgr2/data/request?outputdataset=/WplusToMuNu_NNPDF30_TuneEE5C_13TeV-powheg-herwigpp/RunIISummer15wmLHEGS-MCRUN2_71_V1-v1/GEN-SIM 
This data should not change over time
Also please make sure that we return the same data with the same format compared to the old version.
Check the old version over: https://cmsweb.cern.ch/couchdb/reqmgr_workload_cache/_design/ReqMgr/_view/byoutputdataset?key=%22/WplusToMuNu_NNPDF30_TuneEE5C_13TeV-powheg-herwigpp/RunIISummer15wmLHEGS-MCRUN2_71_V1-v1/GEN-SIM%22 

3. getWorkflowById: 
Check for https://cmsweb.cern.ch/reqmgr2/data/request?prep_id=task_TOP-RunIIFall17wmLHEGS-00462 
This response should not change over time
Similarly compare the result w/ couchdb call in the previous version. 

4. getReqmgrInfo: 
Check for https://cmsweb.cern.ch/reqmgr2/data/info But note that the values in this response will change, perhaps you can only check for keys.

5. getAgentConfig
Agent names and configurations always change, so I think it's not good idea to use one VALID response of today as a long-term test data. For now, you can locally test your implementation w/ the following example call, but let's do not introduce a test for this in the code. We can note it down this as `NOT TESTED` https://cmsweb.cern.ch/reqmgr2/data/wmagentconfig/cmsgwms-submit7.fnal.gov  

6. getWorkflowsByName
This can be tested w/ the output of getWorkflowsByStatus function below, because it's used this way in Unified modules:
https://github.com/CMSCompOps/WmAgentScripts/blob/master/Unified/batchor.py#L24 

7. getWorkflows / getWorkflowsByStatus:
Check for https://cmsweb.cern.ch/reqmgr2/data/request?status=normal-archived&requestor=sagarwal&request_type=ReReco&detail=false
A note for naming: It does not just return workflows by status, it also returns them by user and request type. Let's keep it as getWorkflowsByStatus but think about a better name in the future.  

8. getSchema: 
Check for https://cmsweb.cern.ch/reqmgr2/data/request?name=pdmvserv_task_BPH-RunIIFall18GS-00350__v1_T_201021_154340_8354

9. getSplittingsNew: 
Check for: https://cmsweb.cern.ch/reqmgr2/data/splitting/pdmvserv_task_BPH-RunIIFall18GS-00350__v1_T_201021_154340_8354 
Also, naming is not good, we should think of something better.

amaltaro:
Hasan

> getWorkflows / getWorkflowsByStatus:
> Check for https://cmsweb.cern.ch/reqmgr2/data/request?status=normal-archived&requestor=sagarwal&request_type=ReReco&detail=false

please **never** query for archived workflows, specially for testing purposes. On our backend, this RESTful call is mapped to multiple CouchDB calls, and the first one would return an avalanche of data.

haozturk:
Thanks for the notice @amaltaro I know that without filters it returns a bulk of data and I thought it's safe to use it with filters since it only returns 2 requests with the call above. But, sure we can stop querying this endpoint for archived workflows. Perhaps a rate limiting can be implemented for such endpoints in case clients are not aware of the price of the API.

haozturk:
Here are some further comments:
1. Function names can be more descriptive:
    1. `getWorkflow`: Not clear what it returns. `getWorkflowSchema` is a better name
    2. `getWorkflowByOutput` --> `getWorkflowSchemasByOutput`
    3. `getWorkflowById` --> `getWorkflowByPrepId`

I also started this wiki page for the documentation of the migration. We can create a table there and show the differences for each migrated function/class (Two columns: old version, new version?). This can be used while we're migrating the code which invokes such functions: https://github.com/CMSCompOps/WmAgentScripts/wiki/Python3-Migration-Notes Amanda, I guess, you need some permissions to edit this wiki. I'll take care of that. We can use some other documentation tool as well if you find something more convenient for you.

amanda-grp:
Thanks for the comments!

About confirming the outputs of getWorkflowByOutput/getWorkflowById, they are the same to the old version except when details is True. In this case the format is the same, but there are two missing keys in the dictionaries (_rev and _attachments); the other key/values are all the same too.

> 2\. getWorkflowByOutput:
>     Check for https://cmsweb.cern.ch/reqmgr2/data/request?outputdataset=/WplusToMuNu_NNPDF30_TuneEE5C_13TeV-powheg-herwigpp/RunIISummer15wmLHEGS-MCRUN2_71_V1-v1/GEN-SIM
>     This data should not change over time
>     Also please make sure that we return the same data with the same format compared to the old version.
>     Check the old version over: https://cmsweb.cern.ch/couchdb/reqmgr_workload_cache/_design/ReqMgr/_view/byoutputdataset?key=%22/WplusToMuNu_NNPDF30_TuneEE5C_13TeV-powheg-herwigpp/RunIISummer15wmLHEGS-MCRUN2_71_V1-v1/GEN-SIM%22

src/python/Services/ReqMgr/ReqMgrReader.py
```diff
@@ -6,53 +6,383 @@
...
+    def _runWithRetries(
+        self,
+        fcn: Callable,
+        fcn_pargs: list,
+        fcn_args: dict = {},
+        default: Optional[Any] = None,
+        tries: int = 10,
+        wait: int = 5,
+    ) -> Any:
+        # TODO: is there a better place to put this function?
```
haozturk, line 57: How about in a new module called `RetryManager.py` under `src/python/Utils/`? It's for sure that it should not be in ReqMgrReader.

amaltaro: My experience is that the system became much more robust since it was initially implemented. So I'd be inclined to drop these HTTP retry call logic, it will make the code more clear and maintainable.

src/python/Services/ReqMgr/ReqMgrReader.py
```diff
@@ -6,53 +6,383 @@
...
-        except Exception as e:
+
+        except Exception as error:
             msg = "Error initializing ReqMgrReader\n"
-            msg += "%s\n" % format(e)
+            msg += f"{error}\n"
             raise Exception(msg)
 
-    def getWorkflowByCampaign(self, campaign, details=False):
+    def _runWithRetries(
```
haozturk, line 48: This is a general comment: Let's drop `runWithRetries` in all functions. It really makes the code quite long and unreadable. Once the code goes into production, when we can enable this feature for unstable endpoints. Ideally, it would have been great if we can just list the functions for which you want to retry and all functions are run with retry policy when they are invoked. Anyhow, this is something to think about in the future. For now, let's drop it. We can still keep the function in the Utils package as I mentioned above in case it's needed.

src/python/Services/ReqMgr/ReqMgrReader.py
```diff
@@ -6,53 +6,383 @@
...
+        if make_copy:
+            data = copy.deepcopy(result["result"][0][wf])
+        else:
+            data = result["result"][0][wf]
+            
+        if drop_null:
+            return {k: v for k, v in data.items() if v not in [None, "None"]}
+        return data
+
+    def getWorkflowSchemaByCampaign(
```
haozturk, line 146: Change name to `getWorkflowsByCampaign`. `Schema` is confusing in case `details` is False.

src/python/Services/ReqMgr/ReqMgrReader.py
```diff
@@ -6,53 +6,383 @@
...
+        try:
+            return self._getWorkflowSchemaByParam({"campaign": campaign}, details)
 
+        except Exception as error:
+            self.logger.error(
+                f"Failed to get workflows from reqmgr for campaign {campaign}"
+            )
+            self.logger.error(str(error))
+
+    def getWorkflowSchemaByOutput(
```
haozturk, line 164: Change name to `getWorkflowsByOutput`. `Schema` is confusing in case `details` is False.

src/python/Services/ReqMgr/ReqMgrReader.py
```diff
@@ -6,53 +6,383 @@
...
-                                 param={"campaign": campaign, "detail": str(details)})
+            return self._getWorkflowSchemaByParam({"outputdataset": dataset}, details)
 
+        except Exception as error:
+            self.logger.error(
+                f"Failed to get workflows from reqmgr for dataset {dataset}"
+            )
+            self.logger.error(str(error))
+
+    def getWorkflowSchemaByPrepId(
```
haozturk, line 182: Change name to `getWorkflowsByPrepId`. `Schema` is confusing in case `details` is False.

src/python/Services/ReqMgr/ReqMgrReader.py
```diff
@@ -6,53 +6,383 @@
...
+                time.sleep(wait)
+
+        if default:
+            return default
+        raise Exception("NoDefaultValue")
+
+    def getWorkflowSchema(
+        self,
+        wf: str,
+        drop_null: bool = False,
```
haozturk, line 86: I would remove this argument, I do not see a use case of it. 

Another general comment: 
1. We should avoid a function having too many arguments. It's not a good practice. If a function needs to take more than 2 arguments, we should ask ourselves:
    1. Are we sure that this function is atomic enough? A function should be responsible for just one thing. 
    2. If the answer is yes, keep all arguments under a dict (called **kwargs, or params)

haozturk: If we can achieve this, we can keep the function signature in one line. It would be much readable.

src/python/Services/ReqMgr/ReqMgrReader.py
```diff
@@ -6,53 +6,383 @@
...
+                wait=wait,
+            )
+
+        except Exception as error:
+            self.logger.error(
+                f"Failed to get workflows from reqmgr for {status} (user={user}, type={rtype}, priority={priority})"
+            )
+            self.logger.error(str(error))
+
+    def getWorkflowSchemaByName(
```
haozturk, line 244: Change name to`getWorkflowsByNames`. So that we understand that it gets multiple names and returns multiple workflows.

src/python/Services/ReqMgr/ReqMgrReader.py
```diff
@@ -6,53 +6,383 @@
...
+                {"details": details},
+                tries=tries,
+                wait=wait,
+            )
+
+        except Exception as error:
+            self.logger.error(f"Failed to get workflows from reqmgr for {names}")
+            self.logger.error(str(error))
+
+    def _getWorkflowSchemaByParam(
```
haozturk, line 272: Change name to `getWorkflowsByParam`

haozturk:
Btw, Amanda, I guess it's better to update https://github.com/CMSCompOps/WmAgentScripts/wiki/Python3-Migration-Notes when the PR is merged, so that you won't need to keep updating it after every change.

src/python/Services/ReqMgr/ReqMgrReader.py
```diff
@@ -6,53 +6,383 @@
...
+    ) -> None:
+        # TODO: url is not being used
         try:
             configurationHandler = ConfigurationHandler()
-            self.reqmgrUrl = os.getenv('REQMGR_URL', configurationHandler.get('reqmgr_url'))
+            self.reqmgrUrl = os.getenv(
+                "REQMGR_URL", configurationHandler.get("reqmgr_url")
+            )
+            self.reqmgrEndpoint = {
+                "req": "/reqmgr2/data/request/",
```
haozturk, line 33: `req` to `request`

src/python/Services/ReqMgr/ReqMgrReader.py
```diff
@@ -6,53 +6,383 @@
...
+                return self._getWorkflowSchema(
+                    wf, drop_null=drop_null, make_copy=make_copy, from_cache=True
+                )
+
+    def _getWorkflowSchema(
+        self,
+        wf: str,
+        drop_null: bool = False,
+        make_copy: bool = False,
+        from_cache: bool = False,
```
haozturk, line 123: I would also remove this `from_cache` parameter. Let's do not apply caching for now. We can handle it later.

src/python/Services/ReqMgr/ReqMgrReader.py
haozturk: yes! Totally agree: https://github.com/CMSCompOps/WmAgentScripts/pull/859/files#r658569903

haozturk:
To summarize my first round of review:
1. Make function and variable names as descriptive as possible
2. Keep functions as simple as possible:
    1. Don't use too many arguments
3. Drop retries

src/python/Services/ReqMgr/ReqMgrReader.py
```diff
@@ -6,53 +6,383 @@
...
-            self.reqmgrUrl = os.getenv('REQMGR_URL', configurationHandler.get('reqmgr_url'))
+            self.reqmgrUrl = os.getenv(
+                "REQMGR_URL", configurationHandler.get("reqmgr_url")
+            )
+            self.reqmgrEndpoint = {
+                "req": "/reqmgr2/data/request/",
+                "info": "/reqmgr2/data/info/",
+                "agentConfig": "/reqmgr2/data/wmagentconfig/",
+                "splitting": "/reqmgr2/data/splitting/",
+                "couchdb": "/couchdb/reqmgr_workload_cache/",
```
amaltaro, line 37: Hasan, this `couchdb` is one example that can be replaced by `/reqmgr2/data/request/`

src/python/Services/ReqMgr/ReqMgrReader.py
amanda-grp: I was thinking of defining `runWithRetries` as a decorator, so it will be easier to use it in the future. What do you think?

haozturk: that makes sense to me.

haozturk:
Thanks a lot @amanda-grp  These look very good, so merging it. Can you please make sure that the updates of this PR are reflected properly in https://github.com/CMSCompOps/WmAgentScripts/wiki/Python3-Migration-Notes

amanda-grp:
> Thanks a lot @amanda-grp These look very good, so merging it. Can you please make sure that the updates of this PR are reflected properly in https://github.com/CMSCompOps/WmAgentScripts/wiki/Python3-Migration-Notes

Thanks @haozturk ! The wiki is updated now

PR #878, opened 2021-07-16
Removing the campaigns

scarletnorberg:
RunIISummer19UL17DIGI
RunIISummer19UL17DIGIPremix
RunIISummer19UL17GEN
RunIISummer19UL17HLT
RunIISummer19UL17pLHE
RunIISummer19UL17pLHEGEN
RunIISummer19UL17RECO
RunIISummer19UL17SIM
RunIISummer19UL17wmLHEGEN
RunIISummer19UL18DIGI
RunIISummer19UL18DIGIPremix
RunIISummer19UL18GEN
RunIISummer19UL18HLT
RunIISummer19UL18pLHE
RunIISummer19UL18pLHEGEN
RunIISummer19UL18RECO
RunIISummer19UL18SIM
RunIISummer19UL18wmLHEGEN
RunIISummer19ULPrePremix

We still have no answer on this do we have ReqMgr status filter in this endpoint? So I did my best to check that wf's were normal archived and not just announced. Something we should still check on after the current fire is put out.


haozturk:
@scarletnorberg  All campaigns are good to close. There is just one problem. One campaign (`RunIISummer19ULPrePremix`) is not closed, although it's listed. Can you please close that one as well?

scarletnorberg:
Yep I forgot let me double check that that was on jordans list do not want to close something that they actually want to keep just because I think it is not needed.

scarletnorberg:
Finally found the original message and it is there so I will update the PR

scarletnorberg:
It is removed

PR #879, opened 2021-07-16
Changing pileup

scarletnorberg:
I removed the secondary locations because I thought with primary AAA it should run fine. I also forgot to open a transfer ticket but have opened one to ask about the placement of the secondary. 
PRCAMPAIGNS-192
PRCAMPAIGNS-193
PRCAMPAIGNS-194
PRCAMPAIGNS-195
PRCAMPAIGNS-196
PRCAMPAIGNS-197
https://its.cern.ch/jira/browse/CMSTRANSF-283

### DIFF
--- a/campaigns.json
+++ b/campaigns.json
@@ -2873,6 +2873,14 @@
     "go": true,
     "lumisize": -1,
     "maxcopies": 1,
+    "SecondaryLocation": [
+       "T1_IT_CNAF_Disk",
+       "T1_US_FNAL_Disk"
+    ],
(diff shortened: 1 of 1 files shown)


haozturk:
I think we'll not place the pileups at CERN in EU. So, while updating secondary locations, please remember to blacklist CERN and HLT from those campaigns @scarletnorberg

scarletnorberg:
Adding in this I thought it was done PRCAMPAIGNS-199. There is no APV campaign for FS.

PR #880, opened 2021-07-19
RunIISummer20UL*pp5TeV for 2017 Adding all in slightly different way

scarletnorberg:
All the campaigns have slightly different settings and we still need a pileup file for DR 
Here are the JIRA tickets:
PRCAMPAIGNS-216
PRCAMPAIGNS-217
PRCAMPAIGNS-218
PRCAMPAIGNS-219
PRCAMPAIGNS-220
I think this should hopefully spark some discussion on what is needed in the campaigns.

### DIFF
--- a/campaigns.json
+++ b/campaigns.json
@@ -2693,14 +2693,30 @@
     "fractionpass": 0.95, 
     "go": true, 
     "lumisize": -1, 
-    "maxcopies": 1, 
-    "resize": "auto"
+    "resize": "auto",
+        "T2_US_Purdue",
+        "T2_US_Caltech",
+        "T2_US_Florida",
+        "T2_US_Nebraska",
+        "T2_US_UCSD",
+        "T2_US_Wisconsin"
+      ]
+    },
+  },
   }, 
   "RunIIpp5Spring18pLHE": {
     "fractionpass": 0.95, 
     "go": true, 
     "lumisize": -1, 
-    "tune": true
+        "T2_US_Purdue",
+        "T2_US_Caltech",
+        "T2_US_Florida",
+        "T2_US_Nebraska",
+        "T2_US_UCSD",
+        "T2_US_Wisconsin"
+      ]
+    },
+  }, 
   }, 
   "RunIIpp5Spring18wmLHEGS": {
     "custodial": "T1_FR_CCIN2P3_MSS", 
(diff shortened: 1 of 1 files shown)


haozturk:
@scarletnorberg  couple of questions and comments: 
1. For `RunIISummer20UL17pp5TeVwmLHEGS` how do you know LFN base should be `/store/himc`?
2. We don't need `force-complete` parameter. `fractionpass` already gives us the information we need. Unified closes the workflows if output is > `fractionpass` . So, let's remove that parameter.
3. It's okay to remove `maxcopies`. It's only used in MSOutput and the default is already 1. https://github.com/dmwm/WMCore/wiki/ReqMgr2-MicroService-Output#determining-rule-attributes 
4. Let's remove `custodial`. It's not needed anymore
5. `tune` is not needed, either. So, let's remove that one, as well.
6. For blacklist, I do not see any justification. Do you have? If not, I'd suggest removing it as well.

scarletnorberg:
This one we can merge last the reason I did it this way is because of some of the other campaigns around it. This may require a verbal chat.

scarletnorberg:
For 1. https://github.com/CMSCompOps/WmAgentScripts/blob/master/campaigns.json#L2618 this campaign is similar so why not have the same himc?

scarletnorberg:
For 2 great does that mean we can remove it from all campaigns that are currently in the system?

scarletnorberg:
3. Great can we do another PR where we remove it from all campaigns currently in the system?

scarletnorberg:
For 4 and 5 great lets remove it form everything in the system.

scarletnorberg:
For 6 if this is pp like the other one then it is heavy ion and we should. I agree I do not think it is heavy ion and I do not think the other campaign is either the one I link above https://github.com/CMSCompOps/WmAgentScripts/blob/master/campaigns.json#L2618

haozturk:
Scarlet, we agreed on the lfn base. For updating other campaigns for redundant parameters, I agree we should do it. But, let's do it for these campaigns for now. For the rest, we can do it in a different PR. So, in short, for this PR:
1. Update LFN bases
2. Remove `force-complete`, `maxcopies`, `tune` and `custodial`
I'll review the PR again once you made the changes

PR #899, opened 2021-08-10
Migrate siteInfo and remaining functions in utils.py to Python3

amanda-grp:
Fixes #893 

#### Status
In development.
Passing all unit tests.

#### Description
Migration of remaining utils.py functions to Python3:
- `utils.py/siteInfo` to WorkflowMgmt/SiteController.py
- `utils.py/duplicateAnalyzer` to WorkflowMgmt/DuplicateFilesAnalyzer.py
- `utils.py/userLock` to WorkflowMgmt/UserLockChecker.py
- `utils.py/lockInfo` + `UnifiedLock` to WorkflowMgmt/GlobalLockController.py
- `utils.py/componentCheck` + `componentInfo` to Services/ServicesChecker.py
- `utils.py/forceComplete` to WorkflowMgmt/WorkflowStatusEnforcer.py/WorkflowStatusEnforcer.forceComplete
- `utils.py/invalidate` to  to WorkflowMgmt/WorkflowStatusEnforcer.py/WorkflowStatusEnforcer.invalidate
- `utils.py/get_dashbssb` to Services/MONIT/MONITReader.py/MONITReader.getDashbssb
- `utils.py/getWMStats` to Services/WMStats/WMStatsReader.py/WMStatsReader.getCachedWMStats
- `utils.py/getFailedJobs` to  Services/WMStats/WMStatsReader.py/WMStatsReader.getFailedJobs
- `utils.py/invalidateFiles` to Services/Dynamo/DynamoWriter.py/DynamoWriter.invalidateFiles
- `utils.py/getSiteStorage` to Services/CRIC/CRICReader.py/CRICReader. getSiteStorage
- `utils.py/deep_update` to Utilities/IteratorTools.py/deepUpdate
- `utils.py/checkMemory` to Utilities/ResourceTools.py/checkMemory

This PR also includes:
- The log functions, migrated to `Utilities/Logging.py`
- `reqMgrClient.py/setWorkflowForceComplete` to Services/ReqMgr/ReqMgrWriter.py/ReqMgrWriter.forceCompleteWorkflow
- `reqMgrClient.py/invalidateWorkflow` to to Services/ReqMgr/ReqMgrWriter.py/ReqMgrWriter.invalidateWorkflow

#### To do:
- [ ] The class ServicesChecker depend on the implementation of JIRAClient, McMClient and wtcClient. These should be migrated in a future PR.
- [ ] The class WorkflowController depend on the implementation of McMClient for the migration of notifyRequestor. This should be migrated in a future PR.

[...]

PR #917, opened 2021-08-27
[Py2Py3] Migrate Unified/injector.py to Python3

amanda-grp:
#### Status
not-tested

#### Description
Migration of `Unified/injector.py` to Python3.

The call of `Unified/invalidator.py` was removed and it'll be included in its on cycle.

#### To do:
- [x] The Injector depends on the migration of `Unified/rejector.py`, and `McMClient.py`. This should be done in other PRs before this one can be completed.

#### Related PRs
Depend on the previous PR: #899, #923 and #937

@haozturk


haozturk:
@amanda-grp  Thanks a lot for this wonderful changes. As we discussed, shall we add a testing mode to the module? If this testing mode is enabled, it should only browse/get/process workflows whose names includes `backfill` (case insensitive check). If it's `False`, it should work in production mode. Does that make sense to you? Once we do that and the conflicts are resolved, this PR is good to go. Then we can start our testing. 

Let's add this testing module to every Unified module in such a way that we don't complicate the code.

amanda-grp:
> @amanda-grp Thanks a lot for this wonderful changes. As we discussed, shall we add a testing mode to the module? If this testing mode is enabled, it should only browse/get/process workflows whose names includes `backfill` (case insensitive check). If it's `False`, it should work in production mode. Does that make sense to you? Once we do that and the conflicts are resolved, this PR is good to go. Then we can start our testing.
> 
> Let's add this testing module to every Unified module in such a way that we don't complicate the code.

@haozturk yes, this sounds great! I added the flag and the filter, please check if they are correct.

haozturk:
Thanks a lot @amanda-grp  looks all good! Merging the PR.

PR #938, opened 2021-10-01
Two pilots only succeeded because of the ACDC

scarletnorberg:
Two of these pilots failed. 
https://cms-unified.web.cern.ch/cms-unified/showlog/?search=task_PPD-RunIISpring21UL18FSGSDR-00001
https://cms-unified.web.cern.ch/cms-unified/showlog/?search=task_PPD-RunIISpring21UL16FSGSDR-00001
I asked in the last one the RunIISpring21UL16FSGSD to check and submit another pilot. 
Jordan is now asking for an ACDC and saying he knows the issue and it is fixed but is it?

### DIFF
--- a/campaigns.json
+++ b/campaigns.json
@@ -2822,7 +2822,7 @@
   },
   "RunIISpring21UL18FSGSDR": {
       "fractionpass": 0.95,
-      "go": false,
+      "go": true,
       "lumisize": -1,
       "maxcopies": 1,
       "SiteBlacklist": [
@@ -2915,7 +2915,7 @@
   },
   "RunIISpring21UL17FSGSDR": {
       "fractionpass": 0.95,
-      "go": false,
+      "go": true,
       "lumisize": -1,
       "maxcopies": 1,
       "SiteBlacklist": [
@@ -2995,7 +2995,7 @@
   },
   "RunIISpring21UL16FSGSDR": {
       "fractionpass": 0.95,
-      "go": false,
+      "go": true,
       "lumisize": -1,
       "maxcopies": 1,
       "SiteBlacklist": [


haozturk:
Both of the pilots have failed since the given memory in the original workflow was not enough. So, if the real workflows are going to be submitted w/ this config, they will all fail. Thus, we should make sure that PdmV submits them w/ the proper config. I already talked to Jordan about this. If we are sure that they will inject the workflows w/ proper memory, we can enable the campaign. @scarletnorberg  Can you make a clear statement on this on the JIRA ticket of these campaigns.

You've enabled 3 campaigns, but linked only 2 pilots. What's the pilot of the 3rd one?

scarletnorberg:
The third one is successful

scarletnorberg:
It was reinjected https://dmytro.web.cern.ch/dmytro/cmsprodmon/workflows.php?prep_id=task_PPD-RunIISpring21UL17FSGSDR-00001

haozturk:
Thanks @scarletnorberg  Before merging this PR, I'd like to make sure that PdmV will inject the workflows w/ proper memory configuration. Can you get a clear confirmation from them on JIRA tickets on this matter? (I already talked to them today and they are aware of the situation)

scarletnorberg:
Jordan did say on one JIRA ticket this: "Hi Scarlet Rachel Norberg Hasan Ozturk, could you kindly run recovery to increase memory, please? This is because we injected with 1core/2GB and for FS it is standard to go a little bit higher in memory consumption." If we take this as assent to yes we know the issue and we fixed it. We know in 17 they did then it is already confirmed. Is this enough for us or would you like me to ask specifically Hey is this really fixed?

haozturk:
Let's get an explicit number from them, i.e. let's make sure with which configuration they are going to submit these workflows.

scarletnorberg:
I figured I would just update here:
PRCAMPAIGNS-225 adding pileup Neutrino sample
Some are already successful and go but I wanted to put everything together
PRCAMPAIGNS-235 pilot successful
PRCAMPAIGNS-234 pilot successful
PRCAMPAIGNS-233 pilot successful
PRCAMPAIGNS-232 pilot successful
PRCAMPAIGNS-231 pilot successful
PRCAMPAIGNS-230 pilot successful
PRCAMPAIGNS-196 pilot successful
PRCAMPAIGNS-194 pilot successful
PRCAMPAIGNS-193 pilot successful
PRCAMPAIGNS-192 pilot successful

PR #961, opened 2021-12-01
change stepchain timeperevent calculation

z4027163:
Change the way of calculation time/event of stepchain. So far it does the same factorization on generator task.


Unified/rejector.py
```diff
@@ -211,6 +211,17 @@ def rejector(url, specific, options=None):
                     ## transform the schema into StepChain schema
                     print "Transforming a TaskChain into a StepChain"
                     mcore = 0
+                    ## find maximum mcore
+                    it=1
+                    while True:
```
haozturk, line 216: This is an infinite loop..

haozturk: Fixed it

PR #993, opened 2022-03-17
Neutrino to secondarylocation Minbias to SiteWhitelist

scarletnorberg:
All I did was move around information that was in the file. Every pileup already had a site weather it is at that site or not is a different question. 
I also changed all neutrino to SecondaryLocation and minbias to SiteWhitelist

### DIFF
--- a/campaigns.json
+++ b/campaigns.json
@@ -746,7 +746,7 @@
     "secondary_AAA": true,        
     "secondaries": {
           "/Neutrino_E-10_gun/Run3Summer21PrePremix-120X_mcRun3_2021_realistic_v6-v2/PREMIX": { 
-              "SiteWhitelist": [
+              "SecondaryLocation": [
                   "T1_US_FNAL_Disk",
                   "T1_DE_KIT_Disk"
                       ]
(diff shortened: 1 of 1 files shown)


campaigns.json
```diff
@@ -1139,7 +1139,11 @@
     }, 
     "resize": "auto", 
     "secondaries": {
-      "/Neutrino_E-10_gun/RunIISummer17PrePremix-MCv2_correctPU_94X_mc2017_realistic_v9-v1/GEN-SIM-DIGI-RAW": {}
+      "/Neutrino_E-10_gun/RunIISummer17PrePremix-MCv2_correctPU_94X_mc2017_realistic_v9-v1/GEN-SIM-DIGI-RAW": {
+          "SecondaryLocation": [
+                    "T1_US_FNAL_Disk"
+                        ],
```
haozturk, line 1145: There shouldn't be a comma here. I see the same issue in the following changes as well

scarletnorberg: Yep I agree. I have now run jq . which checks the formatting. I hope that the SecondaryLocation placement is ok.

PR #996, opened 2022-03-23
Pilot successful

scarletnorberg:
[PRCAMPAIGNS-263](https://its.cern.ch/jira/browse/PRCAMPAIGNS-263) pilot successful


scarletnorberg:
We should figure out why it says disallowed should I just post on the computing mattermost?

PR #999, opened 2022-03-30
Mar30th

scarletnorberg:
[PRCAMPAIGNS-279](https://its.cern.ch/jira/browse/PRCAMPAIGNS-279) We may want to blacklist CERN I quess we should check the chain because these are LHE and have to run at CERN but the pileup is not at CERN


scarletnorberg:
[PRCAMPAIGNS-250](https://its.cern.ch/jira/browse/PRCAMPAIGNS-250) pilot successful

PR #1005, opened 2022-04-06
Adding secondary_AAA: true for 4 campaigns

scarletnorberg:
The new FS and the Run3 campaign had no secondaryAAA true so adding that for 4 campaigns

### DIFF
--- a/campaigns.json
+++ b/campaigns.json
@@ -949,7 +949,7 @@
                       "T2_CH_CERN"
                                     ]
         }
-    }
+    },
     "secondary_AAA": true
   },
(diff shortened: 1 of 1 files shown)


scarletnorberg:
RunIISpring22UL18FSwmLHEGSPremix, RunIISpring22UL17FSwmLHEGSPremix, RunIISpring22UL16FSwmLHEGSPremix, Run3Winter22DRPremix

scarletnorberg:
I added a few minbais that have secondary_AAA was not included so included it.

PR #1022, opened 2022-05-30
Changing three campaigns to true

scarletnorberg:
This looks like a lumi miss match but other wise good turning it to true? https://cms-unified.web.cern.ch/cms-unified/showlog/?search=task_PPD-Run3Winter22CosmicGS-00001

This one same thing https://dmytro.web.cern.ch/dmytro/cmsprodmon/workflows.php?prep_id=task_PPD-Run3Winter22CosmicGS-00001

[PRCAMPAIGNS-292](https://its.cern.ch/jira/browse/PRCAMPAIGNS-292)
[PRCAMPAIGNS-291](https://its.cern.ch/jira/browse/PRCAMPAIGNS-291)

[PRCAMPAIGNS-287](https://its.cern.ch/jira/browse/PRCAMPAIGNS-287)This pilot is 100% successful.

### DIFF
--- a/campaigns.json
+++ b/campaigns.json
@@ -985,7 +985,7 @@
     "resize": "auto"
   },
   "Run3Summer22NanoAOD" :{
-        "fractionpass": 0.97,
+        "fractionpass": 0.95,
         "go": true
   },
   "Run3Winter22CosmicGS":{
--- a/campaigns.json
+++ b/campaigns.json
@@ -986,15 +986,15 @@
   },
   "Run3Summer22NanoAOD" :{
         "fractionpass": 0.97,
-        "go": false
+        "go": true
   },
   "Run3Winter22CosmicGS":{
       "fractionpass": 0.95,
-      "go": false
+      "go": true
   },
   "Run3Winter22CosmicDR":{
       "fractionpass": 0.95,
-      "go": false
+      "go": true
   },
   "RunIIAutumn18DR": {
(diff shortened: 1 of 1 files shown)


campaigns.json
```diff
@@ -986,15 +986,15 @@
   },
   "Run3Summer22NanoAOD" :{
         "fractionpass": 0.97,
```
haozturk, line 988: Why 0.97?

scarletnorberg: The PDMV people always want the NanoAOD to be basically 100 of miniAOD because it is not run in the same chain as Gen-MiniAOD it is run after

haozturk: There are nano campaigns whose fractionpass is .95 and .99. I always prefer the lower one if possible as it'll reduce our manual work. If no failure happens they will complete 100% anyways. This fractionpass defines the threshold below which we're going to do manual recovery. Can you please investigate the possibility of lowering to .95 by contacting PDMV? If it's not okay, we can keep it as .97%

scarletnorberg: Ok will go ask them. To open the campaign I am sure they will say fine.

scarletnorberg: I pushed the update this is holding up other campaigns in here they want.

scarletnorberg: I would point out that maybe we can just get ride of this because we never use it and never want to use it?

PR #1039, opened 2022-07-12
Fix 1032

### LINKED ISSUE #1032: Add xrootd option to makeACDC.py
**Impact of the new feature**
Add xrootd option to makeACDC.py in order to improve the automatized too.

Similar to current 'mem' or 'mcore' options, an 'xrootd' option.

lucalavezzo:
- Fixes #1032
- Added readAMs.py, a script to query tni-test.cern.ch/search/search to find assistance-manual workflows using an exitCode.
- Changed how the site selection is handled in assign.py
- Established a worfklow: readAMs.py to obtain problematic workflows, makeACDCs.py to submit ACDCs, asssign.py to read them, which use lists in .txt files between them to communicate which workflows to act on.


autoACDC/autoACDC.py
```diff
@@ -0,0 +1,501 @@
...
+            lfn = self.options['lfn']
+        elif "MergedLFNBase" in self.schema:
+            lfn = self.schema['MergedLFNBase']
+        elif ancestor_wf and "MergedLFNBase" in ancestor_wf.request:
+            lfn = ancestor_wf.request['MergedLFNBase']
+        else:
+            raise Exception("Can't assign the workflow! Please include workflow lfn using --lfn option.")
+        
+        # activity production by default for taskchains, reprocessing for default by workflows
+        if self.options['activity']:
```
haozturk, line 386: Let's not make this optional. We should look at the original workflow for `Dashboard` and use whatever it uses. Perhaps, if you don't do anything, that's the default behavior.

autoACDC/autoACDC.py
```diff
@@ -0,0 +1,501 @@
...
+            "Dashboard": activity,
+            "ProcessingVersion": procversion,
+            "execute": True,
+            "AcquisitionEra": era,
+            "ProcessingString": procstring,
+            "TrustSitelists": self.options['xrootd'],
+            "TrustPUSitelists": secondary_xrootd
+        }
+
+        if self.options['replica']:
```
haozturk, line 413: @amaltaro  do these parameters have any use nowadays?
1. `NonCustodialSites`
2. `NonCustodialSubType`
3. `AutoApproveSubscriptionSites`

I feel like we should just drop those. What do you think?

lucalavezzo:
superseded by #1119

PR #1055, opened 2022-08-16
Fix 1054; autoACDC w/ new data format

lucalavezzo:
- https://github.com/CMSCompOps/WmAgentScripts/pull/1039 already exists to pull autoACDC into the main repo
- this PR is to update autoACDC to the new data format from the rest API
- also added campaign, exitCodeSite, and site as allowed attributes for solutions


haozturk:
Huge thanks to Luca for this! I'm merging this PR. Firstly, we need to fix the file dependencies. One needs to copy some files inside autoACDC file in order to make the script work, which is quite a bad practice.

PR #1072, opened 2022-09-14
Fix 1057

lucalavezzo:
- added xrootd / sceondary xrootd enabling when a random T1 site is chosen (happens when no sites of the requested ones are available)
- added support for any of: `memory = 10000`, `memory = +1000`, `memory = +1000%`. The first sets the memory (as previously supported), the second increases the original request's memory by a value, and the third by a percentage of the original value.
- removed the attributes that are forbidden in the new update of ReqMgr


haozturk:
thanks a lot @lucalavezzo the changes look good to me. I'd prefer to run an autoACDC w/ this version over a small batch before we merge this. Today, I've run a big batch. After it's finalized, I think we can try a small test run.

I have an additional request for this PR. Can you cast the following xrootd options to bool instead of int  
1. https://github.com/lucalavezzo/WmAgentScripts/blob/fix-1057/autoACDC/autoACDC.py#L340
2. https://github.com/lucalavezzo/WmAgentScripts/blob/fix-1057/autoACDC/autoACDC.py#L448-L449

WMCore supports boolean for these params. Providing int might cause issues later on.

lucalavezzo:
@haozturk thank you for the comments; the latest push fixed the xrootd option by enforcing bool type

lucalavezzo:
Also added explicit checkType of inputs mentioned in https://github.com/CMSCompOps/WmAgentScripts/issues/1080

PR #1088, opened 2022-11-02
adding secondary for Commission22DR

tyjyang:
@haozturk for [this](https://its.cern.ch/jira/browse/PRCAMPAIGNS-325) ticket

### DIFF
--- a/campaigns.json
+++ b/campaigns.json
@@ -217,7 +217,16 @@
       "T2_US_MIT",
       "T2_US_Purdue",
       "T2_US_Vanderbilt"
-    ]
+    ],
+    "secondaries": { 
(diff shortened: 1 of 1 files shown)


haozturk:
Hi @tyjyang this is not the secondary that is asked in the ticket. Am I wrong?

And one question: how did you determine the sites? This is a small pileup (~1TB). So, the space shouldn't be a concern. However, I'm not sure about the IO intensity. FNAL should be definitely one of the sites. CERN as well, although we don't have many CPUs available there due to relval and T0 activity. We see too many file issues at RAL, let's not use it. Let's use KIT as a 3rd site in case CERN isn't available. There is no need for a 4th site. It's too many copies for a pileup

tyjyang:
Hi @haozturk, indeed, I copied a minibias input from another campaign and forgot to change the name. Thanks for catching that! I also changed the sites as you suggested.

PR #1110, opened 2022-12-21
remove T2_ES_CIEMAT from Run3 blacklist

z4027163:
Remove T2_ES_CIEMAT from Run3 blacklist, as the testing WF turns out to be success after the retries.


z4027163:
according to latest reply from the thread https://cms-talk.web.cern.ch/t/urgent-el8-release-needs-fix-for-srm-protocol/18705/55

PR #1124, opened 2023-01-25
T2_CH_CERN_HLT is splitting into T2_CH_CERN_HLT and T2_CH_CERN_P5

tyjyang:
@haozturk


tyjyang:
Hi @haozturk, the error is trickier to find than I thought.

So on the jsonlint site, if I add 2 extra "}" at the end of the file, the error would be gone, which I thought indicated that I accidentally added in some "{". However, looking over the git diff, I couldn't locate such changes. Any idea why this could happen?

- Justin

PR #1154, opened 2023-04-06
WIP: MSPileUp Interface

### LINKED ISSUE #1152: MSPileup Interface
**Impact of the new feature**
Pileup placement, campaign management

We need to implement an interface which will read the campaign configuration whenever there is a change and push the changes to MSPileup

Currently, we run the following command whenever we change `campaigns.json`
```
python campaignsConfiguration.py --load campaigns.json
```
Now, I propose that this script should also trigger the pileup updates. 

```
* A given pileup can be a member of multiple campaigns. A pileup should have the same settings in every campaign. Sometimes, operators might make a mistake and don't comply with this rule. We should scan the whole configuration and throw an error and exit if a pileup doesn't have the same settings in every campaign.

* If the previous step succeeds, whenever, we see a pileup in a campaign, we will get the corresponding pileup document from MSPileup
    * If it returns None, create a new pileup document with the following logic:
```
```
{"pileupName": <pileup name seen in the campaign config>,
  "pileupType": <to be discussed at [1]>,
  "expectedRSEs": <to be discussed at [2]>,
  "campaigns": <to be discussed at [3]>,
  "active": <to be discussed at [4]>
}
```

    * Else (it means that pileup is already in place, we need to check if there is any update in its settings)
        * Check the expectedRSEs:
            * If it's not the same, set the expectedRSEs as the list of RSEs defined in the campaign
        * Check the campaigns:
            * If the current campaign name is included in `campaigns`, don't do anything. If not, append it
        * Check the active:
            * If the `active` coming from the microservice is different from what's defined in the campaign, update it according to the campaign config

[1] In Unified, if a pileup includes `minbias`, it's considered as classical mixing, else `premix`. Perhaps, we can use the same logic, however I see pileups such as `/Hydjet_Quenched_Cymbal5Ev8_PbPbMinBias_5020GeV/HiFall15-75X_mcRun2_HeavyIon_v14_75X_mcRun2_HeavyIon_v14-v1/GEN-SIM` in our campaign config and I don't what is its type. @z4027163 is this a pileup which requires heavy read as minbias?

[2] Set the expectedRSEs as the list of RSEs defined in the campaign

[3] A given pileup can be a member of multiple campaigns. We need to append the current campaign name to the already available list of campaigns coming from the microservice if it's not already there.

[4] In Unified campaigns, we have the `go` field which specifies whether a campaign is active or not. If it's true, workflows can pass through assignment, If not, they cannot. We need to decide whether we should set `active` as the same value `go` holds. I think this would work except the following case: Sometimes, we might need to pause a campaign by setting `go: false`, but we might not want to necessarily delete the pileup in this case. In that sense, it makes more sense to me to introduce a new field in our campaign configuration to specify explicitly whether we want to keep the pileup on disk or not. This new field can be named as `active` and it should be defined in the pileup level. @z4027163 @tyjyang what is your opinion on this?

**Additional context**
@amaltaro @z4027163 @tyjyang please let me know what you think of this proposal. I'm planning to work on this tomorrow.

haozturk:
Fixes #1152 

#### Status
in development

#### Description
Currently, I implemented a consistency check on pileups where we exit if 

- if I pileup has different locations in different campaigns
- if there is no location defined for a pileup

Next, I'll work on implementing what I suggested in the attached issue.

#### Related PRs
None

MSPileup APIs

I'll tag people when the PR is mature


haozturk:
The script is ready. MSPileup is in dry-run mode currently. In order to test `campaigns.json` to MSPileup sync, I'll change PU configuration of these pileups to see if the changes are going to be reflected:
1. `/MinBias_TuneCP5_14TeV-pythia8/Phase2Fall22GS-HCalDetIDFix_125X_mcRun4_realistic_v2-v1/GEN-SIM`
Removed, FNAL from secondaries
2. `/MinBias_TuneCP5_13p6TeV-pythia8/Commission22GS-125X_mcRun3_2022_realistic_v3-v1/GEN-SIM`
Added CNAF as a new RSE

I expect MSPileup try to delete FNAL rule and add CNAF rule respectively in dry-run mode.

haozturk:
Okay, dry-run output came out as expected:
```
$ cat reqmgr2ms-20230421-ms-pileup-tasks-8597dfc88-62tmm.log | grep DRY
2023-04-21 09:40:10,251:INFO:MSPileupTasks: DRY-RUN: rule for pileup /MinBias_TuneCP5_13p6TeV-pythia8/Commission22GS-125X_mcRun3_2022_realistic_v3-v1/GEN-SIM and rse T1_IT_CNAF_Disk should have been created
2023-04-21 09:40:10,325:INFO:MSPileupTasks: DRY-RUN: rule id '04d31a1d76784a7d8e434e7f30805402' should have been deleted.
2023-04-21 09:40:10,325:INFO:MSPileupTasks: DRY-RUN: rule id 'f0fe26015a424d61a1bf86634149c030' should have been deleted.
```

PR #1158, opened 2023-04-18
Cleaning up PU not on disk + json format clearnup

tyjyang:
Hi @haozturk, I cleaned up PUs that are no longer on disk. Please check if this is correct, and we should merge and sync before deployment of MSPileup in dry-run mode at 5pm Tuesday CERN time.


haozturk:
Hi @tyjyang thanks a lot.

It's hard to review, because there are lots of cosmetic changes which are related to pileups. Is it possible for you to give me what you have changed regarding pileups?

PR #1163, opened 2023-04-19
change all SecondaryLocation to SiteWhitelist

tyjyang:


tyjyang:
Hi @haozturk, I see that you have pushed the fix for the 3 errors you sent over slack. 

My updated script generated output that are consistent with your version after the error fixes.

In this PR I am unifying the usage of SiteWhitelist and SecondaryLocation in the campaign configuration. This will make our future developments less error-prone. Could you merge this if everything looks good?

@z4027163 FYI

PR #1204, opened 2023-07-25
Add reviveFailed.py script to resubmit/reject failed WF

yanfr0818:
The script check if a failed WF needs resubmission: if it has a younger WF in its family with a valid Unified status, it won't be revived. Currently it doesn't care about which Unified status, so long as the WF has ANY unified status.

The rejection is optional. Use the `-r` option to reject all the failed WFs and invalidate the corresponding output datasets.

### DIFF
--- a/reviveFailed.py
+++ b/reviveFailed.py
@@ -1,0 +1,118 @@
+#!/usr/bin/env python
+import sys
+import os
+import pwd
+import json
+import httplib
+import optparse
+import resubmit, reqMgrClient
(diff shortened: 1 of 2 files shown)


yanfr0818:
@z4027163 I haven't tested the rejection yet. In this script, I intend to reject ALL failed WFs in RegMgr, since they are either revived or being revived after the resubmission. Do you have comments on this?

yanfr0818:
@z4027163 Any update on this?

PR #1209, opened 2023-07-31
handle edge case of overloaded console

lucalavezzo:
Handle the edge case in which the console (or other component) is up, but very slow at responding, AND such component is not necessary (i.e. soft) in componentInfo.


lucalavezzo:
Closing this in favor or replacing soft with ignore

PR #1217, opened 2023-08-09
add back campaign RunIIFall17wmLHEGS

tyjyang:
@z4027163 @jenimal 

We need to add this campaign back for [this](https://dmytro.web.cern.ch/dmytro/cmsprodmon/workflows.php?prep_id=task_B2G-RunIIFall17wmLHEGS-00258) newly injected workflow.


tyjyang:
pdmv informed us to reject the WF instead of re-enabling the campaign

